PR #19, opened 2019-04-17
Remove hello world test

dalg24:


dalg24:
> No, let it live.

Why would you want to keep it?  I only provided it as an example when we move to Boost UTF

PR #20, opened 2019-04-18
Get rid of warnings about redefining BOOST_TEST_MODULE

dalg24:


dalg24:
This gets rid of the warning and starts cleaning up the tests.

PR #25, opened 2019-05-02
Fix unused variable warnings

Rombur:


aprokop:
@Rombur Do you not get any boost warnings?

Rombur:
No I don't but I ran without cuda so no `nvcc_wrapper`

PR #56, opened 2019-05-09
Logo

anagainaru:
This is a first draft. I need to get a better understanding of what is needed from the logo to continue. 

Focus on big picture feedback (width to height ratio, text in the middle of the tree (font, size, position), upside down tree, etc) and not on details. Details can be changed later once we set the main features.


aprokop:
ok to test

dalg24:
> ok to test

Making sure the logo does not break the code?

masterleinad:
> Making sure the logo does not break the code?

You never know.:wink:

aprokop:
Thanks @anagainaru! I actually like it a lot! My preference would be:
- get rid of the bottom tree
  But maybe leave its "grass" so that ArborX title is adorned from both top and bottom
- make the remainder closer to square shape
  Does not have to be square, though
- make the electronic-like components not extend outside of the rectangular box posed by the tree
  This may require removing one of the branches
- Maybe do something special about the X letter
  There is a very nice symmetry in the font between starting A and final R letters, would be nice to somehow emphasize that

aprokop:
> Making sure the logo does not break the code?

yeah, better make sure that the logo does not wreck havoc 🤦‍♂

anagainaru:
> * get rid of the bottom tree
>   But maybe leave its "grass" so that ArborX title is adorned from both top and bottom

Without the bottom tree nobody will understand what the roots are, but agree that it needs to be more discreet. I'll rethink the tree.

> * make the electronic-like components not extend outside of the rectangular box posed by the tree
>   This may require removing one of the branches
> * Maybe do something special about the X letter
>   There is a very nice symmetry in the font between starting A and final R letters, would be nice to somehow emphasize that
> * Maybe do something interesting with the end points of circuits?
>   Maybe organize them in some parallel structure?

I'll see how many elements can be included before it becomes too crowded, but I like the enthusiasm :)
 
> In general, there is also a bit of concern about having a vectorized image for scaling, but I guess if we had a very high density raster image, that could also work.

The final version will be vectorized (not latex, but you are welcome to find a way to convert it from corelDRAW to whatever you like) 

> yeah, better make sure that the logo does not wreck havoc

Security first, good to see my taxes are not wasted

aprokop:
The electronic components look nicer in the second one, but honestly, I like the first one a lot better. And maybe emphasizing X was a stupid idea on my part :|

anagainaru:
> The electronic components look nicer in the second one, but honestly, I like the first one a lot better. And maybe emphasizing X was a stupid idea on my part :|

Good I agree, I like the first one more also :) I can put the electronic from the second on the first one

aprokop:
Like the upper part, but still have the same comments about the bottom part. I'm going to step back now and allow others to comment on the design.

anagainaru:
@aprokop So you want the logo to be an upside down tree without the tree? If I remove the bottom part, the logo will look like an electrocuted squid :)

The bottom part is smaller so the logo is less vertical now and should fit on the github square space. I think you guys should either keep the tree or throw away both the top and bottom (and ask for something else)

aprokop:
> If I remove the bottom part, the logo will look like an electrocuted squid :)

Beauty in the eye of beholder ;)

I personally lean towards something like this:
Screen Shot 2019-05-14 at 5 52 03 PM

masterleinad:
> I personally lean towards something like this:

I also prefer something like that to `logo_v2.0.jpg` but at least parts of the trunk should be visible.

Rombur:
I like v1.1. @aprokop to me your version looks like a bush not the root of tree.

anagainaru:
When the logo is smaller it's even harder to understand what the roots are. With v1.1 you at least get the idea even if you don't see details.

arborx_github2

arborx_github

aprokop:
@dalg24 @Rombur Please weigh in with your opinions.

anagainaru:
If you really don't want the tree, I'd prefer removing all traces of it (and just leave circuits). Something like this:

arborx_v3 0

masterleinad:
So far I like v1.1 best. I find the large suggestion pretty hard to understand.

aprokop:
Lets resurrect this and just decide on one.

masterleinad:
Do you have a new opinion?

aprokop:
Two of you liked v1.1, @dalg24 never told us anything. So lets just go with it.

aprokop:
@dalg24 Do you have an opinion?

aprokop:
Cool. If all agree on 1.1, what's the next step for this? @anagainaru to provide the raw files? 

Of note, I tried changing the ArborX organization's logo to v1.1, but it does not work well because it requires a square (and the width is fixed to that of the image).

anagainaru:
logo_v1_1

This is the _v1_1 I have on my laptop, which has padding to be square, it should be fine for the organization logo. Is this the one you want? I can vectorize it this weekend.

aprokop:
@anagainaru Great! Thanks a lot! 

All, I set th organization profile picture to the latest one. The funny thing, in your personal profile you will see it (the tree), however when you go to the organization page, it still shoes the regular github one. Not sure what's going on.

aprokop:
May also want a transparent variant for profile.

aprokop:
@dalg24 See, you can learn.

masterleinad:
> @dalg24 See, you can learn.

Are you referring to the logo already being visible?

aprokop:
> Are you referring to the logo already being visible?

Not sure I understand.

aprokop:
ping

aprokop:
I think the only remaining thing here is to have the SVG versions. I tried making them myself from the png files of this PR, but was not super successful. I used Inkscape, `Path->Trace bitmap` with the following settings

settings1
settings2

The result is attached (renamed from svg to txt)
[logo_v1_1_nobg.txt](https://github.com/arborx/ArborX/files/4395322/logo_v1_1_nobg.txt)

It does not follow curves well. @anagainaru thank you once again for your work. In one of your previous comments you wrote
> I can vectorize it this weekend.

Were you thinking of a different approach?

anagainaru:
I was planing on using CorelDraw to generate the SVG file. It's not an automatic process so I need to dedicate some time to it. Ask me again Monday if I don't upload it.

aprokop:
@anagainaru Thank you. I don't want to waste your time. I think it would be sufficient if you could briefly write about how it could be done.

anagainaru:
Finished the SVG version, since it's a bit different than the original one (it's impossible to get exactly the same), I'm uploading both the SVG and the corresponding PNG. Take a look and make sure everything looks good.

aprokop:
@anagainaru This looks great! Thank you so much.

Lets merge it after doing the following:
- remove v1.1 versions
- add 1.2_nobg version
- rebase on top of current master
- move logos to `docs/logos`
- update `README.md` to use the logo
- (optional) squash the original commits

@anagainaru I can take care of this if you are OK with it.

anagainaru:
> @anagainaru I can take care of this if you are OK with it.

Sure

@aprokop  let me know if you need help removing the bg

aprokop:
Alright, this should be ready for merge.

aprokop:
For the record: to produce transparent background, I used GIMP:
- `Layer->Transparency->Add Alpha Channel`
- `Edit->Transparency->Color To Alpha`

PR #69, opened 2019-05-31
Enable comparison of tuples in unit tests

dalg24:
Admittedly not the necessarily the best place to put the code that print the tuple...


masterleinad:
I think having it in that file is fine.

PR #82, opened 2019-06-14
Enable C++14 when building Kokkos in automated CI builds on Jenkins

dalg24:
We require C++14 standard but we rely on Kokkos to get the proper compiler flags.
This PR ensures that the `-std=c++14` flag will be present on the compile line in the automated builds on Jenkins.

### DIFF
--- a/.jenkins
+++ b/.jenkins
@@ -15,7 +15,7 @@ pipeline {
                         dockerfile {
                             filename "Dockerfile"
                             dir "docker"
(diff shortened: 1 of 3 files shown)


dalg24:
To be clear this PR does not attempt to solve the problem with are having with Kokkos and CMake regarding the proper way of setting the C++ standard.  I realize now the original title was misleading.

masterleinad:
Would you mind to actually test that this works by using `std::make_unique` in `examples/distributed_tree_driver/distributed_tree_driver.cpp`?

dalg24:
> Would you mind to actually test that this works by using `std::make_unique` in `examples/distributed_tree_driver/distributed_tree_driver.cpp`?

Done.

masterleinad:
:smile:

PR #88, opened 2019-07-09
Fix a few clang-tidy complaints

masterleinad:
Running `clang-tidy` really didn't yield much, but this still looks nicer.


masterleinad:
```
warning: use auto when initializing with a cast to avoid duplicating the type name [modernize-use-auto]
```

PR #89, opened 2019-07-10
Add .clang-tidy file

masterleinad:
Requested in https://github.com/arborx/ArborX/pull/88#pullrequestreview-259707119.

I started from enabling all the `modernize`, `mpi`, `performance` and `readability` checks and disabled the ones that gave warnings I didn't like. The final list of enabled checks is
```
    modernize-avoid-bind
    modernize-deprecated-headers
    modernize-loop-convert
    modernize-make-shared
    modernize-make-unique
    modernize-raw-string-literal
    modernize-redundant-void-arg
    modernize-replace-auto-ptr
    modernize-replace-random-shuffle
    modernize-return-braced-init-list
    modernize-shrink-to-fit
    modernize-unary-static-assert
    modernize-use-auto
    modernize-use-bool-literals
    modernize-use-default-member-init
    modernize-use-emplace
    modernize-use-equals-default
    modernize-use-equals-delete
    modernize-use-noexcept
    modernize-use-nullptr
    modernize-use-override
    modernize-use-transparent-functors
    modernize-use-using
    mpi-buffer-deref
    mpi-type-mismatch
    performance-faster-string-find
    performance-for-range-copy
    performance-implicit-conversion-in-loop
    performance-inefficient-algorithm
    performance-inefficient-vector-operation
    performance-move-const-arg
    performance-move-constructor-init
    performance-noexcept-move-constructor
    performance-type-promotion-in-math-fn
    performance-unnecessary-copy-initialization
    performance-unnecessary-value-param
    readability-avoid-const-params-in-decls
    readability-container-size-empty
    readability-delete-null-pointer
    readability-deleted-default
    readability-else-after-return
    readability-function-size
    readability-identifier-naming
    readability-implicit-bool-conversion
    readability-inconsistent-declaration-parameter-name
    readability-misleading-indentation
    readability-misplaced-array-index
    readability-non-const-parameter
    readability-redundant-control-flow
    readability-redundant-declaration
    readability-redundant-function-ptr-dereference
    readability-redundant-member-init
    readability-redundant-smartptr-get
    readability-redundant-string-cstr
    readability-redundant-string-init
    readability-simplify-boolean-expr
    readability-static-accessed-through-instance
    readability-static-definition-in-anonymous-namespace
    readability-uniqueptr-delete-release
```


.clang-tidy
```diff
@@ -0,0 +1,3 @@
+---
+Checks:          '-*,performance-*,-performance-inefficient-string-concatenation,mpi-*,modernize-*,-modernize-pass-by-value,readability-*,-readability-braces-around-statements,-readability-named-parameter'
```
dalg24: Did you use `--dump-config`?

dalg24: (I am asking because I was surprised by the indentation)

masterleinad:
> I don't necessarily agree with all of the disabling but this is good.

I am happy to be convinced to disable less. In the end, we should address all the `clang-tidy` warnings for the checks we enable.

> Can you confirm that this builds without any warning?

I included a commit with all the changes necessary to make `clang-tidy` happy. Some of these are already part of #88.

masterleinad:
Rebased.

PR #90, opened 2019-08-10
Add .mailmap

aprokop:


.mailmap
```diff
@@ -0,0 +1 @@
+Damien L-G <dalg24@gmail.com> dalg24 <dalg24@gmail.com>
```
masterleinad:
```suggestion
Damien L-G <dalg24@gmail.com>
```

aprokop: Cool, did not know about that.

PR #92, opened 2019-08-22
[WIP] Perform user work

Rombur:
This is a first step for #86 There is no point in looking at the implementation yet, it only works in serial and uses boost even if it is not necessary. I just want to discuss about the [interface](https://github.com/arborx/ArborX/compare/master...Rombur:delegate_work?expand=1#diff-103cb0b3c6d4ff80865cc372c70a9c45R65-R71). In particular, do we still want to have `indices`, `offset`, and `ranks` in the interface. They don't need to be there if we don't return them to the user. Personaly I would remove them but I would still provided them to the user's functor so if the user want to save them it would be her job. The other question is how do we want to get the user's data. For now, the user needs to create a vector of `Data` with as many elements as ranks. This allows me to simply do a `all_to_all` and then, the user can just filter the results the way she wants. The advantage is that it's simple for us and for the user. The disadvantage is that if the data is large the memory consumption may be large.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,10 @@ if(ArborX_ENABLE_MPI)
 endif()
 #target_compile_features(ArborX INTERFACE cxx_std_14)
 
+find_package(Boost REQUIRED COMPONENTS mpi serialization)
+target_link_libraries(ArborX INTERFACE Boost::mpi)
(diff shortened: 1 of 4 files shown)


Rombur:
Any feedback on the API?

Rombur:
After talking with @dalg24 We have decided to go in another direction for this.

PR #97, opened 2019-09-15
Constexpr geometric primitives and algorithms

dalg24:


dalg24:
> Looks good. The error is unrelated to the PR

Unfortunately it is :)
```
note: cannot call non-constexpr function "sqrt(double)"
```

dalg24:
That's a shame it was totally working with gcc 7.4.0

Rombur:
> note: cannot call non-constexpr function "sqrt(double)"

Oh I restarted the tester. It was failing with a missing `gpg` key.

Rombur:
Since @masterleinad is our `constexpr` expert, I will give him a chance to review this PR and I'll merge tomorrow.

src/details/ArborX_DetailsAlgorithms.hpp
```diff
@@ -85,7 +85,7 @@ double distance(Point const &a, Point const &b)
 KOKKOS_INLINE_FUNCTION
 double distance(Point const &point, Box const &box)
 {
-  Point projected_point;
+  Point projected_point{};
```
masterleinad: `Point` does not initialize the data it stores? That's unexpected.

test/tstDetailsAlgorithms.cpp
```diff
@@ -14,159 +14,153 @@
 
 #define BOOST_TEST_MODULE DetailsAlgorithms
 
-namespace details = ArborX::Details;
+// NOTE: message is not required any more with C++17
+#define STATIC_ASSERT(cond) static_assert(cond, "");
```
masterleinad: I would argue that defining a macro is more confusing than typing the extra four characters, but since it's just a test...

test/tstDetailsAlgorithms.cpp
```diff
@@ -14,159 +14,153 @@
...
 BOOST_AUTO_TEST_CASE(distance)
 {
-  BOOST_TEST(details::distance({{1.0, 2.0, 3.0}}, {{1.0, 1.0, 1.0}}) ==
-             std::sqrt(5.0));
+  using ArborX::Details::distance;
+  BOOST_TEST(distance({{1.0, 2.0, 3.0}}, {{1.0, 1.0, 1.0}}) == std::sqrt(5.0));
 
   // box is unit cube
-  ArborX::Box box = {{{0.0, 0.0, 0.0}}, {{1.0, 1.0, 1.0}}};
+  constexpr Box box{{{0.0, 0.0, 0.0}}, {{1.0, 1.0, 1.0}}};
```
masterleinad: That's much nicer!

test/tstDetailsAlgorithms.cpp
dalg24: I had called the initial version of the macro `BOOST_TEST_AND_STATIC_ASSERT` because I thought I remembered a Boost.Test issue about unit tests with no assertion somehow reporting failure.  But return code was still zero when I dropped the `BOOST_TEST`.

dalg24:
> Since we require C++14 we don't need to be that careful anyway.

What do you mean?

masterleinad:
> What do you mean?

`constexpr` functions are extremely limited in C++11.

masterleinad:
Looking at https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/ArborX/detail/PR-93/7/pipeline, `clang` has problems with `std::isfinite`. In fact, the standard doesn't require it to be `constexpr`.

PR #99, opened 2019-09-16
Replace ArborX_ENABLE with ARBORX_ENABLE

Rombur:
@dalg24 saw the light for Kokkos and recognized the superiority of CAPS LOCK. If we merge this PR will need to update spack.


aprokop:
> If we merge this PR will need to update spack.

Fair enough. Should be trivial.

I also wonder why in DTK we already had `SET(ARBORX_ENABLE_MPI TRUE)`.

aprokop:
Updating spack is actually a bit non-trivial. We have two versions of arborx there: `0.8-beta`, and `master`. They would need different options. We have two choices:
1. Do `if` statement in `cmake_args` to switch between options
   This is ugly. We'll have to do version comparison too.
2. We could do a new release (`0.8-beta2`?) and replace old release in spack
   
I would prefer the second option. The changes are not backward compatible, so new version is warranted.

dalg24:
3rd option. We don’t merge until Kokkos CMake is more stable.

Rombur:
It's only a beta release, we can replace it.

PR #107, opened 2019-09-25
Add access traits CUDA example 

dalg24:
Demonstrating how one would specialize `ArborX::Details::Traits::Access` to construct a BVH without having to make a `Kokkos::View` of primitives.

Caught some small bugs along the way (call to `Primitives::extent()` and use of `Primitives::execution_space`).  Might open a separate PR for these.

Will look into improving error messages in a later PR.


examples/access_traits/main.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+int main(int argc, char *argv[])
+{
+  Kokkos::initialize(argc, argv);
+
+  int const N = 10000;
+  float *a;
+  float *d_a;
+
+  a = (float *)malloc(sizeof(float) * N);
```
Rombur: Trying to write code that is `C99` compliant?

examples/access_traits/main.cpp
```diff
@@ -0,0 +1,82 @@
...
+  cudaMalloc((void **)&d_a, sizeof(float) * N);
```
Rombur: Why the cast? Never seen anyone doing that.

examples/access_traits/main.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+  for (int i = 0; i < N; ++i)
+  {
+    a[i] = (float)i;
+  }
+  cudaMemcpy(d_a, a, sizeof(float) * N, cudaMemcpyHostToDevice);
+
+  ArborX::BVH<Kokkos::Cuda::device_type> bvh{Data{d_a, d_a, d_a, N}};
+  ArborX::BVH<Kokkos::Cuda::device_type> bvh2{Data2{d_a, d_a, d_a, d_a, N}};
+
```
Rombur: I don't know if you want to put this into this example or another one but we should do a real search not just building a tree.

dalg24: Haven't decided where it will go yet but planning on doing this next.

src/details/ArborX_DetailsConcepts.hpp
```diff
@@ -60,6 +61,45 @@ struct has_centroid<
...
+        std::enable_if_t<Kokkos::is_memory_space<
+            typename Traits::Access<Primitives>::MemorySpace>::value>,
+        std::enable_if_t<std::is_same<typename Traits::Access<Primitives>::Tag,
+                                      PointTag>::value>,
+        std::enable_if_t<std::is_same<decltype(Traits::Access<Primitives>::get(
+                                          std::declval<Primitives const &>(),
+                                          std::declval<int>())),
+                                      Point>::value>,
+        decltype(Traits::Access<Primitives>::size(
+            std::declval<Primitives const &>()))>> : std::true_type
```
masterleinad: It might be rather hard for a user to understand what the requirements are. What about something like https://godbolt.org/z/HHumU1?

Or are you planning on using SFINAE with `has_access_traits`?

examples/access_traits/main.cpp
```diff
@@ -0,0 +1,82 @@
...
+{
+template <>
+struct Access<Data>
+{
+  __device__ inline static Point get(Data const &d, int i)
+  {
+    return {{d.d_x[i], d.d_y[i], d.d_z[i]}};
+  }
+  inline static typename std::size_t size(Data const &d) { return d.N; }
+  using Tag = PointTag;
```
sslattery: Would this tag dictate the return type of `get()`? i.e. `Tag` dispatches an implementation of the BVH which calls `get()` with an expected return type.

examples/access_traits/main.cpp
```diff
@@ -0,0 +1,82 @@
...
+struct Access<Data2>
+{
+  __device__ inline static Box get(Data2 const &d, int i)
+  {
+    Point const p{{d.d_x[i], d.d_y[i], d.d_z[i]}};
+    float const r = d.d_r[i];
+    return {{{p[0] - r, p[1] - r, p[2] - r}}, {{p[0] + r, p[1] + r, p[2] + r}}};
+  }
+  inline static typename std::size_t size(Data2 const &d) { return d.N; }
+  using Tag = BoxTag;
```
sslattery: Yeah OK this example answers my question from above.

examples/access_traits/main.cpp
```diff
@@ -0,0 +1,82 @@
...
+  float *d_x;
+  float *d_y;
+  float *d_z;
+  float *d_r;
+  int N;
+};
+
+namespace ArborX
+{
+namespace Details
```
sslattery: Does a user-defined traits class make sense in the `Details` namespace or should it be elevated into the `ArborX` namespace?

examples/access_traits/main.cpp
sslattery: This is how NVIDIA documents calling `cudaMalloc`

src/ArborX_LinearBVH.hpp
```diff
@@ -125,22 +125,24 @@ template <typename DeviceType>
 template <typename Primitives>
 BoundingVolumeHierarchy<DeviceType>::BoundingVolumeHierarchy(
     Primitives const &primitives)
```
sslattery: What is the reasoning for the term `Primitives` here?

src/ArborX_LinearBVH.hpp
```diff
@@ -125,22 +125,24 @@ template <typename DeviceType>
...
   // FIXME can be relaxed
-  static_assert(
-      Kokkos::is_view<Primitives>::value,
-      "Must pass a view to the bounding volume hierarchy constructor");
-  static_assert(KokkosExt::is_accessible_from<
-                    typename Details::Traits::Access<Primitives>::MemorySpace,
-                    typename device_type::execution_space>::value,
-                "Primitives must be accessible from bounding volume hierarchy "
-                "execution space");
+  // static_assert(
```
sslattery: Did you mean to comment this out?

examples/access_traits/main.cpp
```diff
@@ -58,23 +61,19 @@ int main(int argc, char *argv[])
   Kokkos::initialize(argc, argv);
 
   int const N = 10000;
-  float *a;
   float *d_a;
 
-  a = (float *)malloc(sizeof(float) * N);
-  cudaMalloc((void **)&d_a, sizeof(float) * N);
+  std::vector<float> a(N);
```
masterleinad: Yes, that's better. Of course, you could also use `std::array` but it doesn't really matter for an example.

src/ArborX_LinearBVH.hpp
dalg24: It is a generic way of qualifying the elements you are indexing.  I like it better than "geometrical objects".

src/details/ArborX_DetailsConcepts.hpp
dalg24: See newer version.

src/ArborX_LinearBVH.hpp
```diff
@@ -125,22 +125,33 @@ template <typename DeviceType>
...
+  static_assert(Details::has_geometry_tag<Primitives>::value,
+                "Traits::Access<Primitives> must define 'Tag' member type that "
+                "is either PointTag or BoxTag");
+  static_assert(
+      Details::has_geometry_accessor<Primitives>::value,
+      "Traits::Access<Primitives> must define 'get()' member function with a "
+      "return type that is consistent with the geometry tag");
+  static_assert(
+      Details::has_size_accessor<Primitives>::value,
+      "Traits::Access<Primitives> must define'size()' member function");
```
dalg24: Just in case you guys were frustrated with the fast compile time.

examples/access_traits/main.cpp
dalg24: Fixed in #112

src/ArborX_LinearBVH.hpp
dalg24: See #113

aprokop:
What are waiting on here? Just searching for stuff?

examples/access_traits/example_cuda_access_traits.cpp
```diff
@@ -0,0 +1,90 @@
...
+    std::iota(std::begin(a), std::end(a), 1.0);
+
+    cudaMemcpy(d_a, a.data(), sizeof(float) * N, cudaMemcpyHostToDevice);
+
+    using device_type = Kokkos::Cuda::device_type;
+    ArborX::BVH<device_type> bvh{PointCloud{d_a, d_a, d_a, N}};
+
+    Kokkos::View<int *, device_type> indices("indices", 0);
+    Kokkos::View<int *, device_type> offsets("offsets", 0);
+    bvh.query(NearestToOrigin{5}, indices, offsets);
```
aprokop: I think it would be better to show how to do multiple queries at a time. Would not want a single query to be what user sees first.

examples/access_traits/example_cuda_access_traits.cpp
```diff
@@ -0,0 +1,90 @@
...
+  {
+    constexpr std::size_t N = 1000;
+    std::array<float, N> a;
+
+    float *d_a;
+    cudaMalloc(&d_a, sizeof(float) * N);
+
+    std::iota(std::begin(a), std::end(a), 1.0);
+
+    cudaMemcpy(d_a, a.data(), sizeof(float) * N, cudaMemcpyHostToDevice);
```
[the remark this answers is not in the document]
dalg24: OK

examples/access_traits/example_cuda_access_traits.cpp
```diff
@@ -0,0 +1,90 @@
...
+  {
+    return {{cloud.d_x[i], cloud.d_y[i], cloud.d_z[i]}};
+  }
+  using memory_space = Kokkos::CudaSpace;
+};
+
+template <>
+struct Access<NearestToOrigin, PredicatesTag>
+{
+  inline static std::size_t size(NearestToOrigin const &) { return 1; }
```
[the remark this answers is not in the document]
dalg24: Yes.  https://github.com/arborx/ArborX/blob/7f58a3b093bd2a24ea5e98082d7c4a193a539738/src/details/ArborX_DetailsBoundingVolumeHierarchyImpl.hpp#L97 for instance

examples/access_traits/example_cuda_access_traits.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+int main(int argc, char *argv[])
+{
+  Kokkos::initialize(argc, argv);
+
+  {
+    constexpr std::size_t N = 1000;
+    std::array<float, N> a;
+
+    float *d_a;
```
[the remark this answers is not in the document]
dalg24: Don't care.

docs/traits.md
```diff
@@ -99,29 +99,31 @@ int main(int argc, char *argv[])
 {
   Kokkos::initialize(argc, argv);
 
-  constexpr std::size_t N = 1000;
-  std::array<float, N> a;
+  {
```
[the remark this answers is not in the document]
aprokop: Can you then move them a bit for clarity, so that it looks like
```c++
   Kokkos::initialize(argc, argv); 
   {

```
and
```c++

   }
   Kokkos::finalize();
```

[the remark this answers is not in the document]
dalg24: OK.  Applied to doc as well.

PR #112, opened 2019-09-26
Promote access traits class template to ArborX:: namespace

dalg24:
Per https://github.com/arborx/ArborX/pull/107#discussion_r328293382 we should not require the user to specialize a class template that lives in the `Details::` namespace.
I deliberately did not act on the geometry tags.  They remain in the `Details::` namespace until further discussion.


src/details/ArborX_Traits.hpp
```diff
@@ -59,12 +57,11 @@ struct Access<View, typename std::enable_if<Kokkos::is_view<View>::value &&
 
   static typename View::size_type size(View const &v) { return v.extent(0); }
 
-  using Tag = PointTag;
+  using Tag = Details::PointTag;
```
dalg24: With these changes the user has to pull the geometry tags from the `Details::` namespace.
I will update #107 to reflect these changes after this PR gets merged.
My intention is to discuss our options regarding the geometry tags then.

PR #111, opened 2019-09-26
Temporarily remove concept checks in tree construction

dalg24:
Will restore later today or tomorrow.
This simplifies greatly history when cherry-picking changes.

PR #113, opened 2019-09-26
Improve error messages in BVH constructor and BVH::query()

dalg24:
As promised here https://github.com/arborx/ArborX/pull/111#issue-321706817
restoring the concept checks.

### DIFF
--- a/src/ArborX_LinearBVH.hpp
+++ b/src/ArborX_LinearBVH.hpp
@@ -18,6 +18,7 @@
 #include <ArborX_DetailsKokkosExt.hpp>
 #include <ArborX_DetailsNode.hpp>
 #include <ArborX_DetailsSortUtils.hpp>
+#include <ArborX_DetailsTags.hpp>
 #include <ArborX_DetailsTreeConstruction.hpp>
 #include <ArborX_Traits.hpp>
 
@@ -59,7 +60,7 @@ public:
     // FIXME placeholder for concept check
 
     using Tag =
-        typename Details::TagHelper<Predicates, Traits::PredicatesTag>::type;
+        typename Details::Tag<Details::decay_result_of_get_t<Access>>::type;
     static_assert(std::is_same<Tag, Details::NearestPredicateTag>::value ||
                       std::is_same<Tag, Details::SpatialPredicateTag>::value,
                   "Invalid tag for the predicates");
--- a/src/ArborX_LinearBVH.hpp
+++ b/src/ArborX_LinearBVH.hpp
@@ -57,8 +57,28 @@ public:
   template <typename Predicates, typename... Args>
   void query(Predicates const &predicates, Args &&... args) const
   {
-    // FIXME placeholder for concept check
-
+    using Access = Traits::Access<Predicates, Traits::PredicatesTag>;
+    static_assert(
+        Details::is_complete<Access>::value,
+        "Must specialize 'Traits::Access<Predicates,Traits::PredicatesTag>'");
+    static_assert(
+        Details::has_memory_space<Access>::value,
+        "Traits::Access<Predicates,Traits::PredicatesTag> must define "
+        "'memory_space' member type that is a valid Kokkos memory space");
+    static_assert(
+        KokkosExt::is_accessible_from<
+            typename Access::memory_space,
+            typename device_type::execution_space>::value,
+        "Traits::Access<Predicates,Traits::PredicatesTag>::memory_space must "
+        "be accessible from the bounding volume hierarchy execution space");
+    static_assert(
+        Details::has_size<Access>::value,
+        "Traits::Access<Predicates,Traits::PredicatesTag> must define "
+        "'size()' member function");
+    static_assert(
+        Details::has_get<Access>::value,
+        "Traits::Access<Predicates,Traits::PredicatesTag> must define 'get()' "
+        "member function");
     using Tag =
         typename Details::Tag<Details::decay_result_of_get_t<Access>>::type;
     static_assert(std::is_same<Tag, Details::NearestPredicateTag>::value ||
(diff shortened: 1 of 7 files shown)


src/details/ArborX_DetailsConcepts.hpp
```diff
@@ -60,6 +61,84 @@ struct has_centroid<
...
+};
+
+template <typename Primitives>
+struct has_size_accessor<
+    Primitives,
+    std::enable_if_t<std::is_integral<decltype(Traits::Access<Primitives>::size(
+        std::declval<Primitives const &>()))>::value>> : std::true_type
+{
+};
+
```
dalg24: In principle these could be templated on the access traits directly.  I don't have a strong opinion on this.  It might be subject to change when we introduce access traits for the predicates.

src/ArborX_LinearBVH.hpp
```diff
@@ -132,7 +132,26 @@ BoundingVolumeHierarchy<DeviceType>::BoundingVolumeHierarchy(
...
+  static_assert(Details::has_geometry_tag<Primitives>::value,
+                "Traits::Access<Primitives> must define 'Tag' member type that "
+                "is either PointTag or BoxTag");
+  static_assert(
+      Details::has_geometry_accessor<Primitives>::value,
+      "Traits::Access<Primitives> must define 'get()' member function with a "
+      "return type that is consistent with the geometry tag");
+  static_assert(
+      Details::has_size_accessor<Primitives>::value,
+      "Traits::Access<Primitives> must define'size()' member function");
```
dalg24:
```suggestion
      "Traits::Access<Primitives> must define 'size()' member function");
```

dalg24:
Ready for review/merge

src/details/ArborX_DetailsConcepts.hpp
```diff
@@ -60,6 +61,91 @@ struct has_centroid<
 {
 };
 
+// is_complete implementation taken from https://stackoverflow.com/a/44229779
+template <class T, std::size_t = sizeof(T)>
+std::true_type is_complete_impl(T *);
+
+std::false_type is_complete_impl(...);
```
sslattery: is the `...` just a place holder for any type? C++14?

src/details/ArborX_DetailsConcepts.hpp
```diff
@@ -60,6 +61,91 @@ struct has_centroid<
...
+};
+
+template <typename T, typename TTag>
+struct has_access_traits<
+    T, TTag, std::enable_if_t<is_complete<Traits::Access<T, TTag>>::value>>
+    : std::true_type
+{
+};
+
+template <typename T, typename = void>
```
sslattery: So this trick with the extra void parameter lets the compiler fall back on this if it can't compile the single parameter version?

src/details/ArborX_DetailsConcepts.hpp
dalg24: It works in C

dalg24:
I didn't want to but I need to add unit tests to demonstrate these actually work...

src/details/ArborX_DetailsConcepts.hpp
```diff
@@ -60,6 +61,91 @@ struct has_centroid<
...
+template <typename T>
+struct has_memory_space<
+    T,
+    std::enable_if_t<Kokkos::is_memory_space<typename T::memory_space>::value>>
+    : std::true_type
+{
+};
+
+template <typename T>
+struct first_template_parameter;
```
sslattery: Could you have done this with `std::tuple` to extract the type at position 0?

dalg24: Yes

src/details/ArborX_DetailsConcepts.hpp
masterleinad: This is one of the usual TMP tricks. The ellipsis is always is the worst candidate in overload resolution. Due to SFINAE, the second overload is only selected if `sizeof(T)` is invalid.

sslattery: makes sense

test/tstDetailsConcepts.cpp
```diff
@@ -0,0 +1,85 @@
...
+    T, std::enable_if_t<!std::is_integral<T>::value>>
+{
+};
+
+static_assert(is_complete<NotSpecializedForIntegralTypes<float>>::value);
+static_assert(!is_complete<NotSpecializedForIntegralTypes<int>>::value);
+
+using ArborX::Details::first_template_parameter_t;
+static_assert(std::is_same<first_template_parameter_t<std::tuple<int, float>>,
+                           int>::value,
```
dalg24: Demonstrate https://github.com/arborx/ArborX/pull/113/files#r331204924 for @sslattery

dalg24:
Added (partial) testing to demonstrate type support added works as intended.

src/details/ArborX_DetailsConcepts.hpp
masterleinad: If we only use it on `Traits::Access` I would certainly prefer that.

src/details/ArborX_Traits.hpp
```diff
@@ -31,10 +31,9 @@ struct PredicatesTag
 {
 };
 
+// Only a declaration so that existence of a specialization can be detected
 template <typename T, typename Tag, typename Enable = void>
-struct Access
-{
-};
+struct Access;
```
masterleinad: Clever!

src/details/ArborX_DetailsConcepts.hpp
dalg24: You are commenting on an old revision.

PR #144, opened 2019-10-12
Replace size_t sort by unsigned int one

aprokop:


aprokop:
This needs a lot more investigation, so a draft for now.

For TIOGA, this is always not worse, and sometimes a lot better.
mesh1_setup
mesh1_search
mesh2_setup
mesh_2_search

src/details/ArborX_DetailsSortUtils.hpp
```diff
@@ -56,13 +57,19 @@ sortObjects(Kokkos::View<unsigned int *, DeviceType> view)
...
   bin_sort.create_permute_vector();
   bin_sort.sort(view);
   ExecutionSpace().fence();
 
-  return bin_sort.get_permute_vector();
+  auto permute_vector = bin_sort.get_permute_vector();
+  Kokkos::parallel_for(
+      ARBORX_MARK_REGION("copy_permute_vector"),
+      Kokkos::RangePolicy<ExecutionSpace>(0, n),
+      KOKKOS_LAMBDA(int i) { permute(i) = permute_vector(i); });
```
dalg24: This is silly now change the return type of the function and skip the copy.

aprokop: That was my first thought too, but we have to make permute indices compatible with `Node*` for storing, so no cigar.

dalg24: Remove https://github.com/arborx/ArborX/blob/dcc40adcc63f6bc253ec31f6050dd969f6e366c2/src/details/ArborX_DetailsTreeConstruction.hpp#L322-L325

I was planning on dropping that static assertion soon anyway.

masterleinad:
So the point is that radix-sort depends on the number type?

aprokop:
@masterleinad I don't think this is a radix sort.

src/details/ArborX_DetailsSortUtils.hpp
aprokop: Not sure about this. It's already a potentially wrong thing to do to work with int. With your proposed change this will spread through the code.

aprokop:
This may not be the best way to go. See #145.

PR #159, opened 2019-10-21
Fix memory space for exclusivePrefixSum

masterleinad:
Without this patch, we try to assign between a `Kokkos::View` with memory space `CudaSpace` and `CudaUVMSpace` in case `Kokkos` is built with `UVM` support and the original memory space is `CudaSpace`. 

Up to now, we tried to use the default memory space associated with the execution space in `ExclusiveScanFunctor` which turned out to be `CudaUVMSpace` in this case. Hence, the constructor call assigns a `Kokkos::View` with memory space `CudaSpace` to a `KokkosView` with memory space `CudaUVMSpace` which is not allowed.

### DIFF
--- a/src/details/ArborX_DetailsUtils.hpp
+++ b/src/details/ArborX_DetailsUtils.hpp
@@ -80,12 +80,13 @@ void exclusivePrefixSum(Kokkos::View<ST, SP...> const &src,
   using ExecutionSpace =
       typename Kokkos::ViewTraits<DT, DP...>::execution_space;
   using ValueType = typename Kokkos::ViewTraits<DT, DP...>::value_type;
+  using DeviceType = typename Kokkos::ViewTraits<DT, DP...>::device_type;
 
   auto const n = src.extent(0);
   ARBORX_ASSERT(n == dst.extent(0));
   Kokkos::parallel_scan(
       "exclusive_scan", Kokkos::RangePolicy<ExecutionSpace>(0, n),
-      Details::ExclusiveScanFunctor<ValueType, ExecutionSpace>(src, dst));
+      Details::ExclusiveScanFunctor<ValueType, DeviceType>(src, dst));
 }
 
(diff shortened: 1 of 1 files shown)


src/details/ArborX_DetailsUtils.hpp
```diff
@@ -80,12 +80,13 @@ void exclusivePrefixSum(Kokkos::View<ST, SP...> const &src,
...
       typename Kokkos::ViewTraits<DT, DP...>::execution_space;
   using ValueType = typename Kokkos::ViewTraits<DT, DP...>::value_type;
+  using MemorySpace = typename Kokkos::ViewTraits<DT, DP...>::memory_space;
 
   auto const n = src.extent(0);
   ARBORX_ASSERT(n == dst.extent(0));
   Kokkos::parallel_scan(
       "exclusive_scan", Kokkos::RangePolicy<ExecutionSpace>(0, n),
-      Details::ExclusiveScanFunctor<ValueType, ExecutionSpace>(src, dst));
+      Details::ExclusiveScanFunctor<ValueType, MemorySpace>(src, dst));
```
dalg24:
```suggestion
      Details::ExclusiveScanFunctor<ValueType, DeviceType>(src, dst));
```

masterleinad: Except that `DeviceType` is also not declared. Do you feel strongly about using `DeviceType` or `MemorySpace` as second template argument? For me, it just looks more appropriate to use a memory space even if you could also pass a device type.

dalg24: You're right sorry.  I prefer to stick with the device type for consistency with the rest unless you have a good reason not to.

masterleinad: I don't feel like fighting for it. I changed to `DeviceType`.

PR #160, opened 2019-10-22
Use agent none at the top-level of the pipeline

dalg24:


dalg24:
"master thread" takes up a slot for nothing.

PR #162, opened 2019-10-24
CUDA-aware MPI

masterleinad:
I still need to clean this up. In particular, the preprocessor variables are not going to stay, but will be replaced by a `CMake` option. So far, I was testing this with
- 60 MPI processes, 
- 10^7 points/MPI process,
- 10^6 queries/MPI process,
- and CUDA

using the `distributed_tree_driver` benchmark.

|  10        | 2.4e0 |  3.7e0 |
| 20         | 4.9e0 | 5.8e0  |
| 40         | 1.0e1 | 1.0e1  |
| 80         | 2.2e1 | 2.0e1  |
| 100        | 2.8e1 | 2.5e1  |
| 120        | 3.3e1 | 3.0e1  |
| 140        | 4.0e1 | 3.4e1  |
| 160        | 4.6e1 | 3.9e1  |


dalg24:
Both CUDA builds were hanging so I ended up killing the build.  @masterleinad please investigate locally and make sure we don't generate traffic jams on the Jenkins server.

masterleinad:
That hopefully works now. I was missing a `cudaDeviceSynchronize` before calling the MPI functions.

src/ArborX_DistributedSearchTree.hpp
```diff
@@ -121,8 +121,9 @@ DistributedSearchTree<DeviceType>::DistributedSearchTree(
   Kokkos::View<Box *, DeviceType> boxes(
       Kokkos::ViewAllocateWithoutInitializing("rank_bounding_boxes"),
       comm_size);
-  // FIXME when we move to MPI with CUDA-aware support, we will not need to
-  // copy from the device to the host
+  // We could avoid copying between device host and device with CUDA-aware MPI,
+  // but this requires the object to be initialized on the device with host data
+  // (_bottom_tree.bounds()). Currently, it doesn't seem worthwhile trying this.
```
dalg24: Try with device data.  Here is how you can get it.
```C++
using TreeAccess = Details::TreeVisualization<DeviceType>::TreeAccess;
TreeAccess::getBoundingVolume(TreeAccess::getRoot(_bottom_tree), _bottom_tree);
```
PS: I am not advocating that is good design but at least it will tell us if this is worth pursuing at all.

dalg24:
image

src/ArborX_DistributedSearchTree.hpp
dalg24: This will be explored in a separate PR.

masterleinad: See #174.

src/details/ArborX_DetailsDistributor.hpp
```diff
@@ -173,6 +186,12 @@ class Distributor
                   _comm, &requests.back());
       }
     }
+
+#ifdef ARBORX_USE_CUDA_AWARE_MPI
+    if (std::is_same<ExecutionSpace, Kokkos::Cuda>::value)
+      cudaDeviceSynchronize();
```
dalg24: Why this  rather than unconditional fence on the execution space?

benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
```diff
@@ -477,7 +477,7 @@ int main(int argc, char *argv[])
     else if (node == "cuda")
     {
 #ifdef KOKKOS_ENABLE_CUDA
-      using Node = Kokkos::Cuda;
+      using Node = Kokkos::Device<Kokkos::Cuda, Kokkos::CudaSpace>;
```
dalg24: This change does not belong to this PR.  At the very least should be its own commit.

masterleinad: It has its own commit now.

masterleinad:
`CUDA-aware MPI` is off by default. Do we want to test for it in one of the configurations?

PR #169, opened 2019-10-28
Enable skipping the build on Jenkins

dalg24:
It does skip the CI but unfortunately it reports failure.  Any thought?


masterleinad:
For me, this doesn't have high priority. At the moment, CI is considerably cheap.

dalg24:
GIT accident again...

PR #179, opened 2019-10-31
Avoid nvcc warning in sortObjects

masterleinad:
This avoids an annoying "dynamic initialization in unreachable code" warning in `sortObjects`.

### DIFF
--- a/src/details/ArborX_DetailsSortUtils.hpp
+++ b/src/details/ArborX_DetailsSortUtils.hpp
@@ -60,9 +60,6 @@ sortObjects(Kokkos::View<unsigned int *, DeviceType> view)
 {
   using ExecutionSpace = typename DeviceType::execution_space;
 
-  static_assert(!std::is_same<ExecutionSpace, Kokkos::Cuda>::value,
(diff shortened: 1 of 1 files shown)


src/details/ArborX_DetailsSortUtils.hpp
```diff
@@ -109,6 +96,27 @@ sortObjects(Kokkos::View<unsigned int *, DeviceType> view)
   return bin_sort.get_permute_vector();
 }
 
+#if defined(KOKKOS_ENABLE_CUDA)
+// NOTE returns the permutation indices **and** sorts the morton codes
+inline Kokkos::View<size_t *, Kokkos::Device<Kokkos::Cuda, Kokkos::CudaSpace>>
```
dalg24: Let the compiler decide whether to inline or not that function.  Drop the `inline` unless you measured?

masterleinad: In this version, the function was not a template, so `inline` is necessary for avoiding ODR problems.

PR #185, opened 2019-10-31
Avoid more compiler warnings for nvcc_wrapper

masterleinad:
This fixes all compiler warnings with `-Wall -Wpedantic -Wextra` using `nvcc_wrapper` for me.


masterleinad:
If that helps to get this merged, here you go.

PR #200, opened 2019-12-05
Allow queries in benchmark to differ from local points

masterleinad:
This is the modification of the `distributed_tree_driver` benchmark used in https://github.com/arborx/ArborX/pull/199#issuecomment-562231451.

### DIFF
--- a/benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
+++ b/benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
@@ -253,10 +253,9 @@ int main_(std::vector<std::string> const &args, const MPI_Comm comm)
                                                                               "local clouds form a board, 3 -> local clouds form a box." )
         ( "do-not-perform-knn-search", "skip kNN search" )
         ( "do-not-perform-radius-search", "skip radius search" )
-        ( "shift-queries" , "Shift the the local box queries are created in from the point cloud. "
-                            "By default, points are reused for the queries. "
-                            "Enabling this option shrinks this box to a third of its size and moves it to the center of the global box. "
-                            "The result is a huge inbalance for the number of queries that need to be processed by each processor.")
+        ( "shift-queries" , "By default, points are reused for the queries. Enabling this option shrinks the local box queries are created "
+                            "in to a third of its size and moves it to the center of the global box. The result is a huge inbalance for the "
+                            "number of queries that need to be processed by each processor.")
         ;
   // clang-format on
   bpo::variables_map vm;
@@ -299,9 +298,9 @@ int main_(std::vector<std::string> const &args, const MPI_Comm comm)
   Kokkos::View<ArborX::Point *, DeviceType> random_points("random_points", 0);
   Kokkos::View<ArborX::Point *, DeviceType> random_queries("random_queries", 0);
   {
-    // Random points are "reused" between building the tree and performing
-    // queries. Note that this means that for the points in the middle of
-    // the local domains there won't be any communication.
+    // By default, random points are "reused" between building the tree and
+    // performing queries. Note that this means that for the points in the
+    // middle of the local domains there won't be any communication.
     Kokkos::resize(random_points, n_values);
     Kokkos::resize(random_queries, n_queries);
 
--- a/benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
+++ b/benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
@@ -361,8 +361,8 @@ int main_(std::vector<std::string> const &args, const MPI_Comm comm)
     auto random_points_host = Kokkos::create_mirror_view(random_points);
 
     // The boxes in which the points are placed have side length two, centered
-    // around offset_[xyz] 2 and scaled by a_values.
-    for (int i = 0; i < n_values; ++i)
+    // around offset_[xyz] and scaled by a_values.
+    for (int i = 0; i < n_construction_values; ++i)
       random_points_host(i) = {{a_values * (offset_x + random()),
                                 a_values * (offset_y + random()),
                                 a_values * (offset_z + random())}};
@@ -386,9 +386,10 @@ int main_(std::vector<std::string> const &args, const MPI_Comm comm)
     else
     {
       // Reuse constructed points both for values and queries.
-      Kokkos::resize(random_queries, n_construction_values);
-      Kokkos::deep_copy(random_queries, random_points_host);
       Kokkos::resize(random_queries, n_queries);
+      Kokkos::deep_copy(random_queries,
+                        Kokkos::subview(random_points_host,
+                                        Kokkos::pair<int, int>(0, n_queries)));
       Kokkos::resize(random_points, n_values);
     }
   }
--- a/benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
+++ b/benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
@@ -253,9 +253,10 @@ int main_(std::vector<std::string> const &args, const MPI_Comm comm)
                                                                               "local clouds form a board, 3 -> local clouds form a box." )
         ( "do-not-perform-knn-search", "skip kNN search" )
         ( "do-not-perform-radius-search", "skip radius search" )
-        ( "shift-queries" , "Shift the queries away from the point cloud. By default, queries and points are generated in the same box "
-                            "on each processor. Enabling this option shrinks this box to a third of its size and moves it to the center "
-                            "of the global box.")
+        ( "shift-queries" , "Shift the the local box queries are created in from the point cloud. "
+                            "By default, points are reused for the queries. "
+                            "Enabling this option shrinks this box to a third of its size and moves it to the center of the global box. "
+                            "The result is a huge inbalance for the number of queries that need to be processed by each processor.")
         ;
   // clang-format on
   bpo::variables_map vm;
--- a/benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
+++ b/benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
@@ -235,6 +235,7 @@ int main_(std::vector<std::string> const &args, const MPI_Comm comm)
   int partition_dim;
   bool perform_knn_search = true;
   bool perform_radius_search = true;
+  bool shift_queries = false;
 
   bpo::options_description desc("Allowed options");
   // clang-format off
@@ -252,6 +253,9 @@ int main_(std::vector<std::string> const &args, const MPI_Comm comm)
                                                                               "local clouds form a board, 3 -> local clouds form a box." )
         ( "do-not-perform-knn-search", "skip kNN search" )
         ( "do-not-perform-radius-search", "skip radius search" )
+        ( "shift-queries" , "Shift the queries away from the point cloud. By default, queries and points are generated in the same box "
+                            "on each processor. Enabling this option shrinks this box to a third of its size and moves it to the center "
+                            "of the global box.")
         ;
   // clang-format on
   bpo::variables_map vm;
@@ -274,6 +278,8 @@ int main_(std::vector<std::string> const &args, const MPI_Comm comm)
     perform_knn_search = false;
   if (vm.count("do-not-perform-radius-search"))
     perform_radius_search = false;
+  if (vm.count("shift-queries"))
+    shift_queries = true;
 
   if (comm_rank == 0)
   {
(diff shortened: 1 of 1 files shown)


dalg24:
Also please provide a rational of we want to add this option in the first place.

masterleinad:
> Also please provide a rational of we want to add this option in the first place.

In terms of communication, this should be similar to the `TIOGA` test case @aprokop's is looking into.

benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
```diff
@@ -252,6 +253,10 @@ int main_(std::vector<std::string> const &args, const MPI_Comm comm)
                                                                               "local clouds form a board, 3 -> local clouds form a box." )
         ( "do-not-perform-knn-search", "skip kNN search" )
         ( "do-not-perform-radius-search", "skip radius search" )
+        ( "shift-queries" , "Shift the the local box queries are created in from the point cloud. "
```
Rombur: I don't understand what the first sentence means. Also there is a typo.

masterleinad: What would help you? Simply deleting the first sentence?

Rombur: Sure but don't forget to change the rest of the comment `...shrinks this box...`

masterleinad:
Some scaling results using 
```
jsrun --smpiargs="-gpu" --nrs 600 --cpu_per_rs 1 --gpu_per_rs 1 --rs_per_host 6 --np 1 --latency_priority CPU-CPU --launch_distribution packed --bind packed:1 ./ArborX_DistributedTree.exe --values=1000000 --queries=100000 --node=cuda --shift=1
```

|  n_procs | knn   |  radius  |
|  1         | 4.30e-1 | 1.73e-1 |
| 2          | 6.47e-1 | 2.37e-1 |
| 4          | 1.05e0 | 3.68e-1 |
| 8        | 1.86e0 | 6.03e-1 |
| 16        | 1.56e0 | 5.49e-1 |
| 32        | 3.32e0 | 1.17e0 |
| 64        | 7.47e0 | 2.71e0 |
| 128        | 1.32e1 | 4.84e0 |
| 256        | 1.57e1 | 5.85e0 |
| 512        | 3.37e1 | 1.34e1 |

benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
```diff
@@ -349,12 +361,23 @@ int main_(std::vector<std::string> const &args, const MPI_Comm comm)
...
-      random_points_host(i) = {
-          {offset_x + random(), offset_y + random(), offset_z + random()}};
+    for (int i = 0; i < n_values; ++i)
+      random_points_host(i) = {{a_values * (offset_x + random()),
+                                a_values * (offset_y + random()),
+                                a_values * (offset_z + random())}};
+    int const max_offset = 2 * shift * i_max;
+    for (int i = 0; i < n_queries; ++i)
+      random_queries_host(i) = {
+          {a_queries * ((offset_x + random()) / 3 + max_offset / 3),
```
aprokop: I *think* this is correct. But could you please add a comment explaining what you do here, and what are the sizes and placement of queries?

benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
```diff
@@ -285,22 +291,26 @@ int main_(std::vector<std::string> const &args, const MPI_Comm comm)
...
+  Kokkos::View<ArborX::Point *, DeviceType> random_queries("random_queries", 0);
   {
-    // Random points are "reused" between building the tree and performing
-    // queries. Note that this means that for the points in the middle of
-    // the local domains there won't be any communication.
-    auto n = std::max(n_values, n_queries);
-    Kokkos::resize(random_points, n);
+    // By default, random points are "reused" between building the tree and
+    // performing queries. Note that this means that for the points in the
+    // middle of the local domains there won't be any communication.
```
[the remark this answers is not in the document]
masterleinad: Without this patch, the locations related to the individual queries coincide with the points (or boxes) we construct the BVH with. Hence, there is no communication required for queries inside the local boxes. This is different after this patch because we typically need to forward most of the queries.

benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
```diff
@@ -285,22 +291,26 @@ int main_(std::vector<std::string> const &args, const MPI_Comm comm)
...
+    Kokkos::resize(random_points, n_values);
+    Kokkos::resize(random_queries, n_queries);
```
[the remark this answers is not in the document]
masterleinad: Can you elaborate? Assignment performs a shallow copy and we only use `n_queries` for the construction. Hence, we should not access uninitialized elements or get a segmentation fault.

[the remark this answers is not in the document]
aprokop: The problem is if you run `--values=100 --queries=10000` it will instead run `--values=100 --queries=100` due to the last assignment, as `random_points` size is 100.

benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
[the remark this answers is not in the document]
aprokop: I don't understand what it means middle here. And won't it depend on the original subdomain shift, too? If you have shift > 0.5, all points will be communicated.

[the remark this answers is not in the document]
masterleinad: Sure, but the default is `shift=1`. If all that is too confusing, I am happy to just remove the comment or replace it with something better if you have a suggestion.

benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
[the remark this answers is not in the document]
masterleinad: Correct, let me revise that.

benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
```diff
@@ -349,10 +349,48 @@ int main_(std::vector<std::string> const &args, const MPI_Comm comm)
...
+        shift_queries ? n_values : std::max(n_values, n_queries);
+
+    // Generate random points uniformly distributed within a box.
+    auto const a_values = std::cbrt(n_construction_values);
+
+    Kokkos::resize(random_points, n_construction_values);
+    auto random_points_host = Kokkos::create_mirror_view(random_points);
+
+    // The boxes in which the points are placed have side length two, centered
+    // around offset_[xyz] 2 and scaled by a_values.
```
aprokop: 2?

benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
```diff
@@ -349,10 +349,48 @@ int main_(std::vector<std::string> const &args, const MPI_Comm comm)
...
+            {a_values * ((offset_x + random()) / 3 + max_offset / 3),
+             a_values * ((offset_y + random()) / 3 + max_offset / 3),
+             a_values * ((offset_z + random()) / 3 + max_offset / 3)}};
+      Kokkos::deep_copy(random_queries, random_queries_host);
+    }
+    else
+    {
+      // Reuse constructed points both for values and queries.
+      Kokkos::resize(random_queries, n_construction_values);
+      Kokkos::deep_copy(random_queries, random_points_host);
```
aprokop:
```suggestion
      Kokkos::deep_copy(random_queries, random_points);
```

benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
```diff
@@ -349,10 +349,48 @@ int main_(std::vector<std::string> const &args, const MPI_Comm comm)
...
+
+    // Generate random points uniformly distributed within a box.
+    auto const a_values = std::cbrt(n_construction_values);
+
+    Kokkos::resize(random_points, n_construction_values);
+    auto random_points_host = Kokkos::create_mirror_view(random_points);
+
+    // The boxes in which the points are placed have side length two, centered
+    // around offset_[xyz] 2 and scaled by a_values.
+    for (int i = 0; i < n_values; ++i)
```
aprokop:
```suggestion
    for (int i = 0; i < n_construction_values; ++i)
```

benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
```diff
@@ -349,20 +351,58 @@ int main_(std::vector<std::string> const &args, const MPI_Comm comm)
     }
     }
 
-    for (int i = 0; i < n; ++i)
-      random_points_host(i) = {
-          {offset_x + random(), offset_y + random(), offset_z + random()}};
+    // Generate random points uniformly distributed within a box.
+    // FIXME: should be fixed to 1D/2D
+    auto const a = std::cbrt(n_values);
```
aprokop: Note this: in the latest commit this was changed from `a = std::cbrt(std::max(n_values, n_queries)`. I think this is more appropriate, and would guarantee the tree does not change with changing number of queries.

benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
```diff
@@ -349,20 +351,58 @@ int main_(std::vector<std::string> const &args, const MPI_Comm comm)
...
+      // For the queries, we shrink the global box by a factor three, and
+      // move it by a third of the global size towards the global center.
+      auto random_queries_host = Kokkos::create_mirror_view(random_queries);
+
+      int const max_offset = 2 * shift * i_max;
+      for (int i = 0; i < n_queries; ++i)
+        random_queries_host(i) = {
+            {a * ((offset_x + random()) / 3 + max_offset / 3),
+             a * ((offset_y + random()) / 3 + max_offset / 3),
+             a * ((offset_z + random()) / 3 + max_offset / 3)}};
```
aprokop: I just noticed that this may not be correct. Lets say you have 5 ranks. Your `i_max = std::ceil(std::cbrt(5)) = 2`. The points in `random_queries` would be correctly placed in the `x and `y` directions. However, their `z` placement would be wrong. For ranks 0-3 `offset_z` would be 0, however `max_offset` would be 4 (for `shift = 1.0`). Thus, I think the queries would stick outside of the values domain, thus resulting in no intersections for a bunch of queries.

masterleinad: I mean it's just a different test case than the previous one. What else would you want the behavior to be?
Also, we should still find neighbors. It's just the radius search that might not result in intersections, isn't?

aprokop: My understanding was that we squeeze down and put queries in the middle of existing values domain. The code was not that.

>  Also, we should still find neighbors. It's just the radius search that might not result in intersections, isn't?

Yes, of course.

masterleinad: > My understanding was that we squeeze down and put queries in the middle of existing values domain. The code was not that.

Yes, I agree that that's the idea and it should work for a sufficiently large of MPI processes in that way. I just don't know what else we want to do if shifting moves queries outside of the values domain. For me, adding a comment that that might happen would be good enough.
What else would you want to do in the case you are describing?

aprokop: Yeah, the case is weird, and I'm not certain about the right treatment. Lets say that we are not going to address it.

aprokop:
> It is not necessary to generate more points. Just apply the shift when you generate the predicates within the traits specialization.

@dalg24 As we allow more queries than values, this is no longer a valid objection. Please re-review or dismiss your review.

benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
```diff
@@ -285,22 +291,16 @@ int main_(std::vector<std::string> const &args, const MPI_Comm comm)
               << "#queries/MPI process    : " << n_queries << '\n'
               << "size of shift           : " << shift << '\n'
               << "dimension               : " << partition_dim << '\n'
+              << "shift-queries           : " << shift_queries << '\n'
               << '\n';
   }
 
-  Kokkos::View<ArborX::Point *, DeviceType> random_points("random_points", 0);
+  Kokkos::View<ArborX::Point *, DeviceType> random_values(
```
dalg24, line 298: Bad name

benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
```diff
@@ -349,20 +351,58 @@ int main_(std::vector<std::string> const &args, const MPI_Comm comm)
...
                        KOKKOS_LAMBDA(int i) {
-                         double const x = random_points(i)[0];
-                         double const y = random_points(i)[1];
-                         double const z = random_points(i)[2];
+                         double const x = random_values(i)[0];
+                         double const y = random_values(i)[1];
+                         double const z = random_values(i)[2];
                          bounding_boxes(i) = {{{x - 1., y - 1., z - 1.}},
                                               {{x + 1., y + 1., z + 1.}}};
                        });
```
dalg24, line 408: Access traits work well for that use case.

benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
```diff
@@ -349,20 +351,58 @@ int main_(std::vector<std::string> const &args, const MPI_Comm comm)
...
+      auto random_queries_host = Kokkos::create_mirror_view(random_queries);
+
+      int const max_offset = 2 * shift * i_max;
+      for (int i = 0; i < n_queries; ++i)
+        random_queries_host(i) = {
+            {a * ((offset_x + random()) / 3 + max_offset / 3),
+             a * ((offset_y + random()) / 3 + max_offset / 3),
+             a * ((offset_z + random()) / 3 + max_offset / 3)}};
+      Kokkos::deep_copy(random_queries, random_queries_host);
+    }
```
dalg24, line 395: Same comment about access traits here.

benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
aprokop: I agree. However, it is aligned with `--values` cmd line argument. Personally, i would change both values and queries cmd line names to something like `--num_points` and `--num_queries` both here and bch_driver.

benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
aprokop: While true, you would replace a two lines by many more. Not really worth it.

benchmarks/distributed_tree_driver/distributed_tree_driver.cpp
aprokop: This would require different access traits than above, so more lines of code. Not sure.

aprokop:
Not addressing traits in this PR, maybe later.

PR #213, opened 2020-02-04
Remove absurdity

dalg24:


masterleinad:
> Are you sure we'll never inherit from it? Why not just remove `throw()`?

If we don't need dynamic polymorphism right now, I don't see a reason not to remove it. This is in `details` so backward-compatibility is not required.

dalg24:
> Are you sure we'll never inherit from it? Why not just remove `throw()`?

`std::exception` <- `std::logic_error` <- `ArborX::SearchException`
`std::exception` has virtual constructor

PR #229, opened 2020-02-21
Enable skipping the build on Jenkins

dalg24:
Re-open #169

PR #233, opened 2020-03-01
Disable skipping CI

masterleinad:


masterleinad:
OK. I don't know how to fix #229 breaking CI. I tried renaming the regex in different ways and also renaming the original `[skip ci]` commit message but nothing apart from disabling the plugin seems to work.

Let's just revert that pull request for now, so any other pull request can be tested.

dalg24:
I added a "Filter branches by commit message" on the Jenkins side
```
Branches whose last commit's message contains (case insensitive) the pattern "[ci skip]" or "[skip ci]" will be ignored.
```
This is so weird, I would have sworn the build for b354bf3 was skipped but was built when I pushed ee5cc92

Anyway let's move on

dalg24:
Changed to `Filter pull requests by commit message`
branches -> pr

masterleinad:
How do you want to proceed here? If I understand correctly, you changed some `jenkins` settings that make the plugin obsolete. Do you want to test this further here or shall we just merge (possibly removing your two commits)?

dalg24:
> How do you want to proceed here? If I understand correctly, you changed some `jenkins` settings that make the plugin obsolete. Do you want to test this further here or shall we just merge (possibly removing your two commits)?

Merge.  As is if you guys are fine with the changes I piggy-backed (my preference) or drop them.

It's more like another way to trigger the same plugin.  And no I don't want to investigate further.

masterleinad:
If nobody objects, I'll merge later today.

PR #255, opened 2020-04-02
Stop pretending we can cherry-pick Kokkos headers

dalg24:
The only headers other than `<Kokkos_Core.hpp>` that may be included specifically are `Kokkos_Macros.hpp` or things like `Kokkos_Pair.hpp` or `Kokkos_Array.hpp`

### DIFF
--- a/src/ArborX_DistributedSearchTree.hpp
+++ b/src/ArborX_DistributedSearchTree.hpp
@@ -17,7 +17,7 @@
 #include <ArborX_DetailsUtils.hpp> // accumulate
 #include <ArborX_LinearBVH.hpp>
 
-#include <Kokkos_View.hpp>
+#include <Kokkos_Core.hpp>
 
 #include <mpi.h>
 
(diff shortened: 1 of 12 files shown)


masterleinad:
What is the rationale for this change? Did anything stop working?

>  I'm not sure if we have a single file included everywhere. If we do, we could move it there and get rid of everywhere else.

We only have `ArborX_Config.hpp` for files that use configuration macro variables.

dalg24:
> Did anything stop working?

Yes my iMac doesn't let me connect anymore

> What is the rationale for this change? 

Did you ever try to compile something like
```C++
#include <Kokkos_View.hpp>
void foo(Kokkos::View<int>) {}
```

This is just straight up wrong, if you want to use `Kokkos::View` you have to include `<Kokkos_Core.hpp>`.

dalg24:
This used to not compile either
```C++
#include <Kokkos_Array.hpp>
{
  Kokkos::Array<int, 3> a;
  return 0;
}
```
Bruno and I fixed it on at least two separate occasion.

`Kokkos::View` is a lost cause.

masterleinad:
> Did you ever try to compile something like
> 
> ```c++
> #include <Kokkos_View.hpp>
> void foo(Kokkos::View<int>) {}
> ```

No, I have not. Didn't expect that to not work.

PR #258, opened 2020-04-03
Get rid of one stray view initialization in forwardQueries()

dalg24:
https://github.com/arborx/ArborX/pull/256#discussion_r403223037

Fixup for #257


dalg24:
Re-test this please

PR #260, opened 2020-04-07
Template distributed search tree on the memory space

dalg24:


src/ArborX_DistributedSearchTree.hpp
```diff
@@ -152,6 +155,28 @@ DistributedSearchTree<DeviceType>::DistributedSearchTree(
...
 }
 
+template <typename DeviceType>
+class DistributedSearchTree<
+    DeviceType, std::enable_if_t<Details::is_device_type<DeviceType>::value>>
+    : public DistributedSearchTree<typename DeviceType::memory_space>
+{
+public:
+  using device_type = DeviceType;
+  DistributedSearchTree() = default;
```
masterleinad, line 165: Is there a reason to allow a default constructed `DistributedSearchTree` now?
The base class doesn't have a default constructor so using it should even not work at all (I tried to be sure).

dalg24: I didn't realize I had not provided a default constructor.  In that case it is deleted anyway.  I can remove if you like.

masterleinad: Yes, I think it is reasonable to expect that explicitly defaulted constructors are usable and I don't see why a default constructed object would make sense here.

dalg24: I removed.  I am not sure if you understand it was without effect.

masterleinad: It's much clearer to me that the object can't be default-constructed.

PR #262, opened 2020-04-09
Update Boost installation process in Docker images

dalg24:
Verify PGP signature (because our failure rate for not being able to download key is not high enough) and do not require to find out the SHA256 of a release.


dalg24:
Retest this please

PR #284, opened 2020-04-25
Disable tests, examples, benchmarks by default

aprokop:
Assumption: most users only care about installing and using ArborX.


aprokop:
retest this please

PR #297, opened 2020-05-08
Address clang-tidy-10 warnings

masterleinad:


masterleinad:
We can wait for #296 and then rebase here before merging, of course.

PR #338, opened 2020-07-10
Drop double underscore prefix in bubble{Down,Up}

dalg24:
The identifiers with a double underscore anywhere are reserved.  I knew I would have to change this one day or the other and kept kicking the can down the road.  I was experimenting with clang-tidy and it choked on it.


dalg24:
Retest this please

masterleinad:
Retest this please.

PR #401, opened 2020-09-30
Drop NDEBUG from HIP CI build

masterleinad:
Let's see if the update also helped to resolve the `NDEBUG` issue.


masterleinad:
Apparently, we still need it here.

PR #386, opened 2020-09-22
Add polytope to explore using it as bounding volume

dalg24:
Fix #387.

PR #364, opened 2020-08-18
Implement stackless tree traversal using escape index (ropes)

aprokop:
Here's a figure to help understanding what ropes are (c/o [1]):

Screen Shot 2020-08-18 at 1 14 19 PM-1

[1] Wang, X., Tang, M., Manocha, D., & Tong, R. (2018). Efficient BVH‐based Collision Detection Scheme with Ordering and Restructuring. In Computer Graphics Forum (Vol. 37, No. 2, pp. 227-237).

PR #403, opened 2020-09-30
Reduce number of places that need to see the node type definition

dalg24:
This is the uncontroversial part of #386 that is needed by #364 
Essentially deducing the node type from the bvh or making it a template parameter depending on the context.

I renamed `BVH::{Node -> node_type}` for consistency with the `BVH::bounding_volume_type` member type and to avoid ambiguities with the actual type.


dalg24:
`<ArborX_DetailsNode.hpp>` header is now only included in `<ArborX_LinearBVH.hpp>` and as a workaround in the tree construction unit test.

aprokop (approved):
Good.

PR #406, opened 2020-10-03
Fix a  label and a profiling region to the new query

aprokop:


dalg24 (approved):

aprokop:
All tests but HIP pass. Merging.

PR #410, opened 2020-10-06
Add plotting script for benchmark

masterleinad:
This is a reduced version of @aprokop's script used for the paper. The intent here is that we can pass any kind of (partial) results produced by `benchmarks/bvh_driver/ArborX_BoundingVolumeHierarchy.exe` and display them graphically. At the moment, we print the mean with some error base indicating a transformed confidence interval for the rate with which points are processed.
Optionally, the actually numbers are also printed. Sample output:
[
plot_newest_errorbars
](url)

Feel free to comment on how to improve it.

### DIFF
--- a/scripts/benchmark.py
+++ b/scripts/benchmark.py
@@ -1,0 +1,101 @@
+#!/usr/bin/env python
+
+import json
+import re
+
+def load_json(filename):
+    with open(filename) as f:
+        try:
+            # Try loading a file using JSON reader
+            data = json.loads(f.read())['benchmarks']
+        except ValueError:
+            # Try loading a file using ASCII reader
+            data = []
+            f.seek(0)   # reset the file as it was messed up above
+            try:
+                for line in f:
+                    # Try to parse only the lines starting with "BM_"
+                    if re.search('^BM_', line) != None:
+                        for timer in ['manual_time', 'manual_time_mean', 'manual_time_median', 'manual_time_stddev']:
+                            m = re.search('(^.*/' + timer + ') .*rate=([0-9.]*)(.*)$', line)
+                            if m != None:
+                                name = m.group(1)
+                                rate = float(m.group(2))
+                                rate_scale = m.group(3)
+                                if rate_scale == 'k/s':
+                                    rate = rate * 1000
+                                elif rate_scale == 'M/s':
+                                    rate = rate * 1000000
+                                else:
+                                    raise Exception('Unknown rate scaling: "' + rate_scale + '"')
+                                data.append({'name' : name, 'rate' : rate})
+                                break
+
+            except RuntimeError as e:
+                raise RuntimeError("Caught an error while parsing on line " + str(lineno) + ": " + e.args[0])
+
+    return data
+
+allowed_geometries = ['filled_box', 'hollow_box']
+allowed_algorithms = ['construction', 'radius_search', 'radius_callback_search', 'knn_search', 'knn_callback_search']
+
+# Geometry is only checked for the source cloud
+# We do not check target cloud geometry as they come in pairs:
+#   filled_box/filled_sphere or hollow_box/hollow_sphere
+def parse_benchmark(json_data, algorithm, implementation, timer_aggregate, geometry, num_radius_passes = 2):
+    if geometry not in allowed_geometries:
+        raise Exception('Unknown geometry: "' + geometry + '"')
+    if algorithm not in allowed_algorithms:
+        raise Exception('Unknown algorithm: "' + algorithm + '"')
+
+    geometries = { 'filled_box' : '0', 'hollow_box' : '1', 'filled_sphere' : '2', 'hollow_sphere' : '3' }
+
+    # Escape () in implementation string
+    implementation = implementation.translate(str.maketrans({"(" : "\(", ")" : "\)", "%" : "\%" }))
+
+    sorted = 1
+
+    # Templates:
+    # BM_construction<ArborX::BVH<_DeviceType_>>/_num_primitives_/_source_point_cloud_type_
+    construction_template = algorithm + '<.*' + implementation + '[^/]*/([^/]*)/' + geometries[geometry] + '/'
+    # BM_knn_search<ArborX::BVH<_DeviceType_>>/_num_primitives_/_num_predicates_/_n_neighbors_/_sort_predicate_/_source_point_cloud_type_/_target_point_cloud_type_
+    knn_template = algorithm + '<.*' + implementation + '[^/]*/([^/]*)/([^/]*)/[^/]*/' + str(sorted) + '/' + geometries[geometry] + '/'
+    # BM_knn_search<ArborX::BVH<_DeviceType_>>/_num_primitives_/_num_predicates_/_n_neighbors_/_sort_predicates_/_buffer_size_/_source_point_cloud_type_/_target_point_cloud_type_
+    # we consider (2P) to be represented by buffer_size = 0
+    radius_template_1P = algorithm + '<.*' + implementation + '[^/]*/([^/]*)/([^/]*)/[^/]*/' + str(sorted) + '/[^0][^/]+/' + geometries[geometry] + '/'
+    radius_template_2P = algorithm + '<.*' + implementation + '[^/]*/([^/]*)/([^/]*)/[^/]*/' + str(sorted) + '/0/' + geometries[geometry] + '/'
+
+    num_primitives = []
+    num_predicates = []
+    rates    = []
+
+    for benchmark in json_data:
+        name = benchmark['name']
+
+        # For Google Benchmark v1.5, one could do
+        #  if benchmark['run_type'] != 'aggregate' or benchmark['aggregate_name'] != timer_aggregate:
+            #  continue
+        if re.search(timer_aggregate, name) == None:
+            continue
+
+        if algorithm == 'construction' and re.search(algorithm, name) != None:
+            m = re.search(construction_template, name)
+            if m != None:
+                num_primitives.append(int(m.group(1)))
+                rates.append(benchmark['rate'])
+
+        elif (algorithm == 'knn_search' or algorithm == 'knn_callback_search') and re.search(algorithm, name) != None:
+            m = re.search(knn_template, name)
+            if m != None:
+                num_primitives.append(int(m.group(1)))
+                num_predicates.append(int(m.group(2)))
+                rates.append(benchmark['rate'])
+
+        elif (algorithm == 'radius_search' or algorithm == 'radius_callback_search') and re.search(algorithm, name) != None:
+            m = re.search(radius_template_2P if num_radius_passes == 2 else radius_template_1P, name)
+            if m != None:
+                num_primitives.append(int(m.group(1)))
+                num_predicates.append(int(m.group(2)))
+                rates.append(benchmark['rate'])
+
+    return num_primitives, num_predicates, rates
--- a/scripts/benchmark_plot.py
+++ b/scripts/benchmark_plot.py
@@ -1,0 +1,199 @@
+#!/usr/bin/env python3
+"""benchmark_plot.py
+
+Usage:
+  benchmark_plot.py -a ALGORITHM [-b BACKENDS...] -g GEOMETRY -i INPUT... [-n] [-e] [-t TITLE] [-o OUTPUT_FILE]
+  benchmark_plot.py (-h | --help)
+
+Options:
+  -h --help                         Show this screen.
+  -a ALGORITHM --algo ALGORITHM     Algorithm ['construction', 'radius_search', 'radius_callback_search', 'knn_search', 'knn_callback_search']
+  -b BACKENDS --backends BACKENDS   Backends to parse [default: all]
+  -g GEOMETRY --geometry GEOMETRY   Geometry (source cloud) ['filled_box', 'hollow_box']
+  -i FILE --input-files=FILE        Input file(s) containing benchmark results in JSON format
+  -n --numbers                      Plot numbers [default: False]
+  -e --errors                       Plot error bars [default: False]
+  -o FILE --output-file=FILE        Output file
+  -t TITLE --title=TITLE            Plot title
+"""
+import matplotlib.pyplot as plt
+import numpy as np
+import math
+import re
+from docopt import docopt
+
+from benchmark import load_json, parse_benchmark, allowed_algorithms, allowed_geometries
+
+def find_available_backends(input_jsons):
+    arborx_backend_regex = "[^<]*<ArborX::BVH<([^>]*)>.*"
+
+    backends = set()
+    for input_json in input_jsons:
+        for benchmark in input_json:
+            search_result = re.search(arborx_backend_regex, benchmark['name'])
+            if search_result != None:
+                backends.add(search_result.group(1))
+    backends = list(backends)
+    backends.sort()
+
+    return backends
+
+class NotMatchingValuesAndPredicatesSizesException(Exception):
+    def __init__(self, message, i, backend):
+        super().__init__(message)
+        self.i = i
+        self.backend = backend
+
+def populate_data(input_jsons, backends):
+    all_num_primitives  = []
+    all_rates = []
+    all_errors = []
+
+    for i in range(len(backends)):
+        # For every backend, find all (num_primitives, rate, error) in all input files
+        backend = backends[i]
+
+        backend_num_primitives = []
+        backend_rates = []
+        backend_errors = []
+
+        implementation = 'ArborX::BVH<' + backend + '>'
+        for i in range(len(input_jsons)):
+            file_num_primitives, file_num_predicates, file_rates  = parse_benchmark(input_jsons[i], algorithm, implementation, 'median', geometry)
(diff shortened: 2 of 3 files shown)


scripts/plot.py
```diff
@@ -0,0 +1,147 @@
...
+def libraries_comparison_rate_figure(input_file, algorithm, geometry, numbers):
+    # Phase 0: parse backends
+    backends = set()
+    with open(input_file) as f:
+        for line in f:
+            search_result = re.search("[^<]*<ArborX::BVH<([^>]*)>.*", line)
+            if search_result != None:
+                backends.add(search_result.group(1))
+    backends = list(backends)
+    backends.sort()
```
aprokop, line 39: I like the automatic way. I also think it should have a manual mode. For example, you may want to get a subset of backends to plot.

scripts/plot.py
```diff
@@ -0,0 +1,147 @@
...
+            ax.annotate('{}'.format(height),
+                        xy=(rect.get_x() + rect.get_width() / 2, height-rect.get_height()/2),
+                        xytext=(0, 3),  # 3 points vertical offset
+                        textcoords="offset points",
+                        ha='center', va='bottom')
+
+    width = 0.6
+    scale = 1/len(backends)
+    for i in range(0, len(backends)):
+        barplot = ax.bar(np.arange(len(unified_num_sources)) + scale*(i-1.5) * width,
```
aprokop, line 88: This is the tricky part. The magic number 1.5 really depends on the number of backends. You would want to make a group of columns centered on the tick.

scripts/plot.py
```diff
@@ -0,0 +1,147 @@
...
+  plot.py -a ALGORITHM -g GEOMETRY -i INPUT... [-d] [-l] [-y] [-n] [-o OUTPUT_FILE]
+  plot.py (-h | --help)
+
+Options:
+  -h --help                         Show this screen.
+  -a ALGORITHM --algo ALGORITHM     Algorithm ['construction', 'knn_search', 'radius_search', 'scaling_table']
+  -g GEOMETRY --geometry GEOMETRY   Geometry (source cloud) ['filled_box', 'hollow_box']
+  -i FILE --input-files=FILE        Input file containing benchmark results
+  -o FILE --output-file=FILE        Output file [default: plot.png]
+  -p PLOT_INDEX --plot=PLOT_INDEX   Plot index
```
aprokop, line 14: What's plot index? Is it used?

scripts/plot.py
```diff
@@ -0,0 +1,147 @@
...
+
+    # Transform given timings into the plotting data
+    data = np.zeros([len(backends), len(unified_num_sources)])
+    errors_data = np.zeros([len(backends), 2, len(unified_num_sources)])
+    for i in range(0, len(backends)):
+        for j in range(0, len(num_sources[i])):
+            index = unified_num_sources.index(num_sources[i][j])
+            data[i, index] = (num_sources[i][j] / timings[i][j])/10**6
+            errors_data[i, 0, index] = num_sources[i][j]/10**6*(1/timings[i][j] - 1/(timings[i][j]+errors[i][j]))
+            errors_data[i, 1, index] = num_sources[i][j]/10**6*(1/(timings[i][j]-errors[i][j]) - 1/timings[i][j])
```
aprokop, line 68: Hmm, this should be fixed.

scripts/plot.py
```diff
@@ -0,0 +1,147 @@
...
+    plt.xticks(fontsize=18)
+    ax.set_xticks(np.arange(len(num_source_labels)))
+    ax.set_xticklabels(num_source_labels)
+    ax.tick_params(axis='x', which='major', bottom='off', top='off')
+    ax.set_xlim([-0.5, len(unified_num_sources)-0.5])
+
+    ax.set_ylabel('Rate (million objects/second)', fontsize=20)
+    plt.yticks(fontsize=18)
+    ax.yaxis.grid('on')
+    plt.title('Construction', fontsize=18)
```
aprokop, line 110: This should not be hardcoded, and instead should be mapped from the `algorithm`.

aprokop (changes requested):
Good start. 

Please make me a co-author (at least, on the first commit).

aprokop:
I also wanted to ask. I assume that CUDA and AMD will be run on different machines, and thus produce different output files. I did not see anything in the script that allows to combine multiple files. Would that have to be done manually prior to running the script?

masterleinad:
> I also wanted to ask. I assume that CUDA and AMD will be run on different machines, and thus produce different output files. I did not see anything in the script that allows to combine multiple files. Would that have to be done manually prior to running the script?

I am not quite sure what all the actual use cases are. So far, I was assuming to just read a single file but it should not be difficult to read multiple if that is more convenient.

scripts/plot.py
masterleinad: What else do you want?

scripts/plot.py
```diff
@@ -0,0 +1,137 @@
...
+
+    # Transform given timings into the plotting data
+    data = np.zeros([len(backends), len(unified_num_sources)])
+    errors_data = np.zeros([len(backends), 2, len(unified_num_sources)])
+    for i in range(0, len(backends)):
+        for j in range(0, len(num_sources[i])):
+            index = unified_num_sources.index(num_sources[i][j])
+            data[i, index] = (num_sources[i][j] / timings[i][j])/10**6
+            errors_data[i, 0, index] = num_sources[i][j]/10**6*(1/timings[i][j] - 1/(timings[i][j]+errors[i][j]))
+            errors_data[i, 1, index] = num_sources[i][j]/10**6*(1/(timings[i][j]-errors[i][j]) - 1/timings[i][j])
```
aprokop, line 64: I think Damien is correct. Looking at the [propagation of uncertainty](https://en.wikipedia.org/wiki/Propagation_of_uncertainty#Example_formulae) for `f = A/B`, the deviation of the inverse `1/B` is indeed `dev(B)/B^2`. So, I think it would be sufficient to do this:
```suggestion
			scaling_factor = num_sources[i][j]/10**6
            data[i, index] = scaling_factor / timings[i][j]
            error = errors[i][j]/(timings[i][j]**2)
            errors_data[i, 0, index] = data[i, index] - scaling_factor * error
            errors_data[i, 1, index] = data[i, index] + scaling_factor * error
```
While it is still not exactly correct (as we are using medians), I think it should be closer. 

P.S. I have not tested this, and it may contain misprints.

aprokop:
I know it may annoy you, but there is one more thing. For people who have trouble distinguishing colors (I know at least couple), it is generally helpful to provide some markings on the bars in addition to colors.

scripts/plot.py
masterleinad: For me the question is how you want to interpret these bars and I normally think of them as some kind of confidence interval. I could be convinced that the original distribution is normal-like but the inverse is certainly not so I don't know how much value providing a standard deviation has.
In practice, the difference should be small anyway. 
If both of you rather want the formula above, tell me again and I will change.

masterleinad:
@masterleinad I think I addressed all your suggestions apart from the error bars now. I am assuming to squash all the commits in the end anyway so only added you as co-author to the last one even if it is not directly related (to avoid force-pushing early).

aprokop:
@masterleinad Thanks! I am not going to argue about errors right now. Lets wait to see if @dalg24 has comments, while i take final look at the rest of the pr.

scripts/plot.py
```diff
@@ -0,0 +1,151 @@
...
+        num_sources.append(tmp_num_sources)
+        timings.append(tmp_timings)
+        errors.append(tmp_errors)
+        unified_num_sources = unified_num_sources + tmp_num_sources
+
+    unified_num_sources = list(set(unified_num_sources))
+    unified_num_sources.sort()
+    num_source_labels = [None] * len(unified_num_sources)
+    for i in range(len(unified_num_sources)):
+        num_source_labels[i] = '$10^' + str(round(math.log(unified_num_sources[i], 10))) + '$'
```
aprokop, line 64: I'm struggling to understand what this does. Could you elaborate on what the end result of `num_sources`, `timings` or `errors` is going to be if there are repetitions in multiple files?

scripts/plot.py
```diff
@@ -0,0 +1,151 @@
...
+
+    # Check input
+    if geometry != 'filled_box' and geometry != 'hollow_box':
+        raise Exception('Unknown geometry: "' + geometry + '"')
+    if algorithm != 'construction' and algorithm != 'knn_search' and algorithm != 'radius_search':
+        raise Exception('Unknown algorithm: "' + algorithm + '"')
+
+    backends = backends.split(',')
+    input_files = input_files[0]
+    input_files = input_files.split(',')
```
aprokop, line 139: I think that if you use docopt's `INPUT...`, you can do `-i file1 -i file2 -i files3` automatically producing a list, so you don't need this.

scripts/plot.py
masterleinad: If there are repetitions only the last run for a specific setting is plotted. This is actually not related to the code above but rather to how we initialize `data` below.

scripts/plot.py
masterleinad: I can try that. Specifying input files via `-i file1,file2,file3` was my first intuition so I tried to make that work. I don't care much one way or the other, though.

masterleinad: Yes, commenting these two lines `-i file1 -i file2 -i files3` works. Do you prefer that syntax?

aprokop: Yes, this way we rely on docopt and no write the code ourselves. Could do the same with backends.

masterleinad: Why would you design interfaces rather for ease of implementation than for ease of use?
Anyway, I don't feel strongly it so I'll just change.

masterleinad: Here you go!

aprokop:
I think what needs to be done here to finally merge it is:
- Switch to parsing JSON files instead of regular output
  It should be more robust, and would simplify the code through removing `benchmark.py`, relying instead on the regular JSON parsing libraries
- Switch to using rates introduced in #457.

@dalg24 You had a piece of code somewhere already to parse benchmark json files already, correct? Can you please share?

aprokop:
Screen Shot 2021-01-11 at 9 34 36 AM

Woah, did not realize that one can also close pull requests automatically from another pull request. Thought it only affected issues.

aprokop:
I updated the benchmark plotting to use rates introduced in #457 and parse JSON files instead of text output. I think this is ready for final review.

aprokop (dismissed):

aprokop:
Aggregate names and runtimes were [added](https://github.com/google/benchmark/commit/58588476ce52c0af4cb54f9c596c4579f78bc953) in v1.5.

aprokop:
One thing that could be nice, but is entirely optional, is to try to parse file exactly once. I'm thinking we could call `populate_data()` and produce lists (including the backends) first, and then trim the lists depending on the `backends` argument. When plotting results from larger runs with multiple file, I noticed that it may takes 10+ seconds for a plot, and couple minutes to cycles through all combinations of algorithms and geometries. Would be nice to bring it down to somewhat interactive levels.

scripts/benchmark_plot.py
```diff
@@ -0,0 +1,179 @@
+#!/usr/bin/env python3
+"""plot.py
+
+Usage:
+  plot.py -a ALGORITHM [-b BACKENDS...] -g GEOMETRY -i INPUT... [-n] [-e] [-o OUTPUT_FILE]
+  plot.py (-h | --help)
```
masterleinad, line 6:
```suggestion
"""benchmark_plot.py

Usage:
  benchmark_plot.py -a ALGORITHM [-b BACKENDS...] -g GEOMETRY -i INPUT... [-n] [-e] [-o OUTPUT_FILE]
  benchmark_plot.py (-h | --help)
```

scripts/benchmark.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python
+
+import json
+import re
+
+def load_json(filename):
+    with open(filename) as f:
+        data = json.loads(f.read())
+    return data
```
masterleinad, line 9: What is the advantage of requiring `json`? IMHO it's easier to just use the plain format since this is created by the benchmark by default.

masterleinad: I checked that it works but I don't necessarily want to create the output on hte same nodes that I am running and copying the much shorter plain format it just easier.

scripts/benchmark_plot.py
```diff
@@ -0,0 +1,179 @@
...
+    input_files = options['--input-files']
+    output_file = options['--output-file']
+    plot_numbers = options['--numbers']
+    plot_errors = options['--errors']
+
+    plt.figure(figsize=(6, 5))
+
+    # Check input
+    if geometry not in ['filled_box', 'hollow_box']:
+        raise Exception('Unknown geometry: "' + geometry + '"')
```
masterleinad, line 165:
```suggestion
        raise Exception('Unknown geometry: "' + geometry + '". Allowed values are "filled_box" and "hollow_box".')
```

scripts/benchmark_plot.py
```diff
@@ -0,0 +1,179 @@
...
+
+    if algorithm not in ['construction', 'knn_search', 'radius_search']:
+        raise Exception('Unknown algorithm: "' + algorithm + '"')
```
masterleinad, line 168:
```suggestion
        raise Exception('Unknown algorithm: "' + algorithm + '". Allowed values are "construction", "knn_search" and "radius_search"')
```

scripts/benchmark.py
aprokop: For me, the advantage of using JSON is that we already have to dump it if we want to use `compare_bench.py` script for run comparison. But you bring a valid point that removing the old "one-line" console output may be stupid. I feel that it brings few points worth discussing:
- With JSON, I believe that we could figure out a way to parse each file exactly once, as we are able to store file as a dictionary in a variable. I think this is challenging with console output
- The rates output is done in human form, using `k` or `M` suffix to indicate thousands/millions. This leads to extra complexity.

This is not to say that I'm opposed to bringing parsing output back. It does have its benefit in its simplicity.

scripts/benchmark_plot.py
aprokop: Allowed values are seen in `./benchmark_plot.py --help`, no need for them here.

scripts/benchmark_plot.py
aprokop: Allowed values are listed with `--help`, no need for them here.

scripts/benchmark_plot.py
masterleinad: I know. Just my user experience: If I misspell an option so I prefer knowing what the correct spelling/allowed options are immediately without needing to relaunch with `--help`.

aprokop:
I think the best way to address reading regular text output is to provide a utility function in plotting that is able to convert it to json. This way, there would be no code overlap with JSON reader, and it would be agnostic to the actual data.

scripts/benchmark.py
aprokop: @masterleinad I added the console output file reader.

aprokop:
OK, this does what I want it to. Still has some rough edges, and certainly places could use some love, but lets try to work on merging it. In the current form, it is extremely useful, and I've been playing with it all day producing figures.

scripts/benchmark_plot.py
masterleinad: Are you fine with me adding this myself in a follow-up pull request?

aprokop: Oh, I forgot about these. Just push to this PR.

scripts/requirements.txt
```diff
@@ -0,0 +1,4 @@
+docopt
+matplotlib
+numpy
+json
```
dalg24, line 4:
```suggestion
```

aprokop: why?

dalg24: because it is a bug.  it is part of the python standard library.  if you try to install the requirements you will get an error

masterleinad (commented):
I'm fine with moving forward here.

aprokop:
@masterleinad You missed couple new options that I added like `radius_callback_search` and `knn_callback_search`. That's why I said we should have a single place with all the options somewhere, so that we won't run into this.

scripts/benchmark.py
```diff
@@ -36,11 +36,16 @@ def load_json(filename):
 
     return data
 
+allowed_geometries = ['filled_box', 'hollow_box']
```
dalg24, line 39: `geometries.keys()`

aprokop: Don't understand.

aprokop: OK, understood now. `requirements.txt` only to contain non-core libraries. Surprisingly to me, `json` is part of core.

scripts/requirements.txt
aprokop: OK, understood now. `requirements.txt` only to contain non-core libraries. Surprisingly to me, `json` is part of core.

aprokop:
@dalg24 Did you have other comments?

aprokop:
I terminated the testing (forgot to add `[ci skip]` to the commit message).

aprokop (approved):

PR #413, opened 2020-10-13
Require Boost 1.56.0 for benchmarks

masterleinad:
Fixes https://github.com/arborx/ArborX/issues/311. This is the first version that works for me.

### DIFF
--- a/benchmarks/CMakeLists.txt
+++ b/benchmarks/CMakeLists.txt
@@ -1,4 +1,4 @@
-find_package(Boost REQUIRED COMPONENTS program_options)
+find_package(Boost 1.56.0 REQUIRED COMPONENTS program_options)
 
 add_subdirectory(bvh_driver)
 if (ARBORX_ENABLE_MPI)


aprokop (approved):

dalg24:
Looking at Boos.Geometry's [changelog](https://www.boost.org/doc/libs/1_68_0/libs/geometry/doc/html/geometry/release_notes.html#geometry.release_notes.boost_1_56) it looks like it is because of our using `is_valid`

masterleinad:
The first error I was seeing was
```
/tmp/ArborX/test/ArborX_BoostRTreeHelpers.hpp:41:36: error: ‘index’ in namespace ‘boost::geometry’ does not name a type
 using Parameter = boost::geometry::index::linear<16>;
                                    ^~~~~
```

dalg24:
The changelog says also in improvements `The rtree pack-creation algorithm optimization (thanks to Lu Wang)` so they might have edited template params

dalg24:
Failure is unrelated.  Merging.

PR #423, opened 2020-11-30
Add CI for SYCL

masterleinad:
With the next Kokkos release most of the tests should work.

PR #441, opened 2020-12-23
Fixup builtin reducer type requirements for SYCL backend 

dalg24:
Amend to #439 because new SYCL build introduced in #423 did not run.

SYCL backend reduction displays similar type requirements than the OpenMPTarget.
The code should still been see as a workaround.  We reported the issue and will address on the Kokkos side.

### DIFF
--- a/src/details/ArborX_Box.hpp
+++ b/src/details/ArborX_Box.hpp
@@ -105,7 +105,7 @@ struct Box
   }
 
 // FIXME Temporary workaround until we clarify requirements on the Kokkos side.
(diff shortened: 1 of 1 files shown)


aprokop (approved):
This PR fixes what it is supposed to, i.e. it allows SYCL backend to build and run tests. The fact that the `ArborX_LinearBVH_Test` fails with the same error as in the `HIP` is out of scope of this PR.

masterleinad:
I have never seen it failing when I was running the tests with the other implementation but I will try to reproduce the problem here.

PR #455, opened 2021-01-08
Allow running BVH benchmark with SYCL and OpenMPTarget explicitly

masterleinad:


aprokop (approved):

dalg24 (approved):
Looks like an oversight in #422

PR #461, opened 2021-01-15
Add ray tracing example

wjge:
The example populates a box with spheres, and shoots rays from the bottom of the box.


examples/raytracing/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+add_executable(ArborX_Raytracing.exe example_raytracing.cpp)
+target_link_libraries(ArborX_Raytracing.exe ArborX::ArborX)
+add_test(NAME ArborX_Raytracing_Example COMMAND ./ArborX_Raytracing.exe)
```
aprokop, line 3: Just a nitpick: `Raytracing` -> `RayTracing`

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,224 @@
+/****************************************************************************
+ * Copyright (c) 2012-2021 by the ArborX authors                            *
```
aprokop, line 2:
```suggestion
 * Copyright (c) 2017-2021 by the ArborX authors                            *
```

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+struct Callback
+{
+  ArborX::Primitives boxes;
+
+  Kokkos::View<float *, MemorySpace, Kokkos::MemoryTraits<Kokkos::Atomic>>
+      accumulator;
+
+  KOKKOS_FUNCTION float dotproduct(ArborX::Point const &v1,
+                                   ArborX::Point const &v2) const
```
aprokop, line 51: You don't do `dotproduct` with `Point`. Use `Vector` here.

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,224 @@
...
+  {
+    return v1[0] * v2[0] + v1[1] * v2[1] + v1[2] * v2[2];
+  }
+
+  KOKKOS_FUNCTION float solveQuadratic(const float &b, const float &c) const
```
aprokop, line 56: The function does not do what its name suggests. It tries to mix solving quadratic with some kind of intersection related things, making it hard to comprehend.

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,224 @@
...
+        if (c > 0)
+          return std::abs(q - c / q);
+        // one point is behind
+        else
+          return q;
+      }
+    }
+  }
+
+  KOKKOS_FUNCTION float overlap(ArborX::Ray const &ray,
```
aprokop, line 90: better call it something like `overlap_distance`, as `overlap` may suggest boolean function.

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+    float b = 2.0 * dotproduct(dir, L);
+    float c = dotproduct(L, L) - r * r;
+
+    return solveQuadratic(b, c);
+  }
+
+  template <typename Predicate>
+  KOKKOS_FUNCTION void operator()(Predicate const &predicate,
+                                  int primitive) const
```
aprokop, line 118: call it `primitive_index`

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,224 @@
...
+  }
+  using memory_space = MemorySpace;
+};
+} // namespace ArborX
+
+struct Callback
+{
+  ArborX::Primitives boxes;
+
+  Kokkos::View<float *, MemorySpace, Kokkos::MemoryTraits<Kokkos::Atomic>>
```
aprokop, line 47: Don't really need atomic here right now, but that's implementation dependent, so doesn't hurt.

aprokop (commented):
A bunch of comments. I have not studied the `Ray`/`Box` intersection algorithm, but I think it can be done simpler.

wjge:
> A bunch of comments. I have not studied the `Ray`/`Box` intersection algorithm, but I think it can be done simpler.

Do you mean Ray/Sphere intersection? The algorithm in the current example can be found here:

https://www.scratchapixel.com/lessons/3d-basic-rendering/minimal-ray-tracer-rendering-simple-shapes/ray-sphere-intersection

examples/raytracing/CMakeLists.txt
wjge: Updated.

aprokop:
> Do you mean Ray/Sphere intersection? The algorithm in the current example can be found here:
> 
> [scratchapixel.com/lessons/3d-basic-rendering/minimal-ray-tracer-rendering-simple-shapes/ray-sphere-intersection](https://www.scratchapixel.com/lessons/3d-basic-rendering/minimal-ray-tracer-rendering-simple-shapes/ray-sphere-intersection)

Your current example has `Ray/Box` intersections, not `Ray/Sphere`.

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,224 @@
...
+#include <ArborX.hpp>
+
+#include <Kokkos_Core.hpp>
+
+#include <iostream>
+#include <random>
+#include <vector>
+
+using ExecutionSpace = Kokkos::DefaultExecutionSpace;
+using MemorySpace = ExecutionSpace::memory_space;
```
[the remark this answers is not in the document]
wjge: I was following the example_callback.
Please suggest 

> > Do you mean Ray/Sphere intersection? The algorithm in the current example can be found here:
> > [scratchapixel.com/lessons/3d-basic-rendering/minimal-ray-tracer-rendering-simple-shapes/ray-sphere-intersection](https://www.scratchapixel.com/lessons/3d-basic-rendering/minimal-ray-tracer-rendering-simple-shapes/ray-sphere-intersection)
> 
> Your current example has `Ray/Box` intersections, not `Ray/Sphere`.

Ray/Box is the bool intersects() implemented in the ArborX::detail, this example is a call back to give the Ray/Sphere intersection length. The box is bounding the sphere.

wjge:
> Your current example has `Ray/Box` intersections, not `Ray/Sphere`.

Ray/Box is the bool intersects() implemented in the ArborX::detail, this example is a call back to give the Ray/Sphere intersection length. The box is bounding the sphere.

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,224 @@
...
+    float x0 = rd_uniform() * L - r;
+    float y0 = rd_uniform() * L - r;
+    float z0 = rd_uniform() * L - r;
+
+    float x1 = x0 + 2.0 * r;
+    float y1 = y0 + 2.0 * r;
+    float z1 = z0 + 2.0 * r;
+
+    return ArborX::Box{{x0, y0, z0}, {x1, y1, z1}};
+  });
```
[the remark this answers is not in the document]
dalg24: `std::vector` is fine.  Use unmanaged view to copy to device w/o having the host intermediate.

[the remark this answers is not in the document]
dalg24:
```C++
Kokkos::deep_copy(boxes_view, Kokkos::View<Box*, Kokkos::HostSpace, Kokkos::MemoryUnmanaged>{boxes.data(), boxes.size()});
```

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,224 @@
...
+#include <random>
+#include <vector>
+
+using ExecutionSpace = Kokkos::DefaultExecutionSpace;
+using MemorySpace = ExecutionSpace::memory_space;
+
+namespace ArborX
+{
+using Primitives = Kokkos::View<Box *, MemorySpace>;
+using Predicates = Kokkos::View<Ray *, MemorySpace>;
```
dalg24, line 26: This is not OK to put random things in ArborX:: namespace.  Specializing access traits is the only legal thing to do.

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,232 @@
...
+    float sinazimuth = std::sin(rd_uniform() * 2.0 * M_PI);
+    float cosazimuth = std::sqrt(1 - sinazimuth * sinazimuth);
+    float cospolar = 1.0 - rd_uniform();
+    float sinpolar = std::sqrt(1.0 - cospolar * cospolar);
+
+    float dirx = sinpolar * cosazimuth;
+    float diry = sinpolar * sinazimuth;
+    float dirz = cospolar;
+
+    rays_host[i] = {{x, y, z}, {dirx, diry, dirz}};
```
aprokop, line 218: No idea what's happening here, please add a comment describing what and how is being constructed.

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,232 @@
...
+  auto rd_uniform = [&]() { return uniform(gen); };
+
+  // radius
+  const float mu_R = 1.0;
+  const float sigma_R = mu_R / 3.0;
+
+  std::normal_distribution<> normal{mu_R, sigma_R};
+  auto rd_normal = [&]() { return std::max(normal(gen), 0.0); };
+
+  Kokkos::View<ArborX::Sphere *, MemorySpace> spheres_view("spheres_view", num_spheres);
```
aprokop, line 190: Drop `_view` postfix from views.

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+    rays_host[i] = {{x, y, z}, {dirx, diry, dirz}};
+  }
+
+  Kokkos::deep_copy(rays_view, rays_host);
+
+  {
+    Predicates<MemorySpace> rays{rays_view};
+    // accumulator
+    Kokkos::View<float *, MemorySpace> overlap_view("overlap_view", num_rays);
```
aprokop, line 226:
```suggestion
    Kokkos::View<float *, MemorySpace> overlap_view("Testing::overlap", num_rays);
```

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,232 @@
...
+  bpo::store(bpo::command_line_parser(argc, argv).options(desc).run(), vm);
+  bpo::notify(vm);
+
+  std::uniform_real_distribution<float> uniform{0.0, 1.0};
+  std::default_random_engine gen;
+  auto rd_uniform = [&]() { return uniform(gen); };
+
+  // radius
+  const float mu_R = 1.0;
+  const float sigma_R = mu_R / 3.0;
```
aprokop, line 185:
```suggestion
  // Random parameters for Gaussian distribution of radii
  const float radius_mu = 1.0;
  const float radius_sigma = mu_R / 3.0;
```

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,232 @@
...
+    }
+  }
+
+  KOKKOS_FUNCTION float overlap_distance(ArborX::Ray const &ray,
+                                         ArborX::Sphere const &sphere) const
+  {
+    auto center = sphere.centroid();
+    auto r = sphere.radius();
+
+    ArborX::Point orig = ray.origin();
```
aprokop, line 124:
```suggestion
    ArborX::Point const &orig = ray.origin();
```

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+    Vector L{orig[0] - center[0], orig[1] - center[1], orig[2] - center[2]};
+
+    //  for normalized direction vector a = 1.0
+    auto dir = ray.direction();
```
aprokop, line 129:
```suggestion
    auto const& dir = ray.direction();
```

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+    float b = 2.0 * dotproduct(dir, L);
+    float c = dotproduct(L, L) - r * r;
+
+    return solveQuadratic(b, c);
```
aprokop, line 134: Please document an algorithm you are using, explaining how you calculate the length.

aprokop: OK, here's how I would document it:
```
Finding the length of the ray intersection with a sphere consists of two steps. First, find two points on a sphere that a bidirectional ray intersects with. These points are the solution to the equation:
   || (ray_origin + alpha * direction) - sphere_center || = radius.
The resulting quadratic equation is
   || direction ||^2 alpha^2 + 2*(ray_origin - sphere_center, direction) alpha + (radius^2 - ||ray_origin - sphere_center||^2) = 0.

If this equation has a solution, ray intersects with a sphere. Once the alpha values are computed, say alpha_1 and alpha_2 (which could be the same if ray just touches the sphere), the directional ray intersection part will be the interval [alpha_1, alpha_2] intersected with [0, +inf). If intersection is empty, directional ray does not intersect. Once that interval is computed, say, [a, b], the length of intersection is simply (b-a)*||direction||.
```

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,232 @@
...
+      }
+      else
+      {
+        // both intersections are AHEAD
+        if (c > 0)
+          return std::abs(q - c / q);
+        // one point is behind
+        else
+          return q;
+      }
```
aprokop, line 114: Are you sure this works when direction is not normalized?

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,231 @@
...
+    return v1[0] * v2[0] + v1[1] * v2[1] + v1[2] * v2[2];
+  }
+
+  KOKKOS_FUNCTION float solveQuadratic(const float &b, const float &c) const
+  {
+    float delta = b * b - 4.0 * c;
+    if (delta <= 0.0)
+      // NO intersection
+      return 0.0;
+    else
```
masterleinad, line 89: readability-else-after-return

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+      // printf("q=%f, c/q = %f \n", q, c / q);
+
+      if (q < 0)
+      {
+        // both intersections are BEHIND
+        if (c > 0)
+          return 0.0;
+        // one point is behind
+        else
```
masterleinad, line 102: readability-else-after-return

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,231 @@
...
+          return c / q;
+      }
+      else
+      {
+        // both intersections are AHEAD
+        if (c > 0)
+          return std::abs(q - c / q);
+        // one point is behind
+        else
```
masterleinad, line 111: readability-else-after-return

aprokop (approved):
It looks good to me. But I worked with @wjge on this PR, so need one other independent review.

aprokop:
@wjge I did not realize that when you said you are going to do a rebase, you planned to merge `master` into this branch. This is not the right way to do it. You should have used `git rebase` command instead. Not asking to change now.

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,300 @@
...
+  {
+    float xi_1 = rand_uniform();
+    float xi_2 = rand_uniform();
+
+    rays_host(i) = {
+        ArborX::Point{rand_uniform() * L, rand_uniform() * L, 0.f},
+        ArborX::Ray::Vector{float(std::cos(2 * M_PI * xi_2) * std::sqrt(xi_1)),
+                            float(std::sin(2 * M_PI * xi_2) * std::sqrt(xi_1)),
+                            std::sqrt(1.f - xi_1)}};
+  }
```
aprokop, line 268: @wjge Now that I think about it, is there a reason to only shoot rays from the bottom of the domain? Why not do that from all the faces of the `[0, L]^3` box? It would give us a better opportunity to investigate sorting of rays.

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,300 @@
...
+  //
+  // The sphere can be expressed as the solution to
+  //     |p - c|^2 - r^2 = 0,           (1)
+  // where c is the center of the sphere, and r is the radius. On the other
+  // hand, any point on a bidirectional ray satisfies
+  //     p = o + t*d,                   (2)
+  // where o is the origin, and d is the direction vector.
+  // Substituting (2) into (1),
+  //     |(o + t*d) - c|^2 - r^2 = 0,   (3)
+  // resulting in a quadratic equation for unknown t
```
masterleinad, line 125:
```suggestion
  // results in a quadratic equation for unknown t
```

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,300 @@
...
+  //     a2 * t^2 + a1 * t + a0 = 0
+  // with
+  //     a2 = |D|^2, a1 = 2*(d, o - c), and a0 = |o - c|^2 - r^2.
+  // Then, we only need to intersect the solution internval [tmin, tmax] with
+  // [0, +inf) of unidirectional ray.
```
masterleinad, line 130:
```suggestion
  // [0, +inf) for the unidirectional ray.
```

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,300 @@
...
+    auto const &c = sphere.centroid();
+    auto const &r = sphere.radius();
+    auto const &o = ray.origin();
+    auto const &d = ray.direction();
+
+    // Vector oc = o - c
+    Vector oc{o[0] - c[0], o[1] - c[1], o[2] - c[2]};
+
+    float a2 = 1.f; // directions are normalized
+    float a1 = 2.0 * dotProduct(d, oc);
```
masterleinad, line 143:
```suggestion
    float const a1 = 2.f * dotProduct(d, oc);
```
for consistency

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,300 @@
...
+  KOKKOS_FUNCTION float overlapDistance(ArborX::Ray const &ray,
+                                        ArborX::Sphere const &sphere) const
+  {
+    auto const &c = sphere.centroid();
+    auto const &r = sphere.radius();
+    auto const &o = ray.origin();
+    auto const &d = ray.direction();
+
+    // Vector oc = o - c
+    Vector oc{o[0] - c[0], o[1] - c[1], o[2] - c[2]};
```
masterleinad, line 140:
```suggestion
    Vector const oc{o[0] - c[0], o[1] - c[1], o[2] - c[2]};
```

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,300 @@
...
+
+    float a2 = 1.f; // directions are normalized
```
masterleinad, line 142:
```suggestion
    float const a2 = 1.f; // directions are normalized
```

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,300 @@
...
+    float a1 = 2.0 * dotProduct(d, oc);
+    float a0 = dotProduct(oc, oc) - r * r;
```
masterleinad, line 144:
```suggestion
    float const a0 = dotProduct(oc, oc) - r * r;
```

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,300 @@
...
+
+    float t1;
+    float t2;
+    if (!solveQuadratic(a2, a1, a0, t1, t2))
+    {
+      // No intersection of a bidirectional ray with the sphere
+      return 0.f;
+    }
+    float tmin = KokkosExt::min(t1, t2);
```
masterleinad, line 153:
```suggestion
    float const tmin = KokkosExt::max(0.f, KokkosExt::min(t1, t2));
```

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,300 @@
...
+    float tmax = KokkosExt::max(t1, t2);
```
masterleinad, line 154:
```suggestion
    float const tmax = KokkosExt::max(t1, t2);
```

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,300 @@
...
+
+    if (tmax < 0)
+    {
+      // Half-ray does not intersect with the sphere
+      return 0.f;
+    }
+
+    // Overlap [tmin, tmax] with [0, +inf)
+    tmin = KokkosExt::max(0.f, tmin);
```
masterleinad, line 163:
```suggestion
```

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,300 @@
...
+
+    // As direction is normalized,
+    //   |(o + tmax*d) - (o + tmin*d)| = tmax - tmin
+    return (tmax - tmin);
+  }
+
+  template <typename Predicate>
+  KOKKOS_FUNCTION void operator()(Predicate const &predicate,
+                                  int primitive_index) const
```
masterleinad, line 172:
```suggestion
                                  int const primitive_index) const
```

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,300 @@
...
+  {
+    auto const &ray = ArborX::getGeometry(predicate);
+    auto const &sphere = _spheres(primitive_index);
+
+    float length = overlapDistance(ray, sphere);
```
masterleinad, line 177:
```suggestion
    float const length = overlapDistance(ray, sphere);
```

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,300 @@
...
+    int i = getData(predicate);
```
masterleinad, line 178:
```suggestion
    int const i = getData(predicate);
```

masterleinad (commented):
Only minor comments.

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,300 @@
...
+      return true;
+    }
+
+    // Instead of doing a simple
+    //   0.5 * (-b +- std::sqrt(discriminant)) / (2*a)
+    // we use a more stable algorithm with less loss of precision (see, for
+    // example, https://www.scratchapixel.com/lessons/3d-basic-rendering/
+    // minimal-ray-tracer-rendering-simple-shapes/ray-sphere-intersection).
+    auto sgn = [](float x) { return (x > 0) - (x < 0); };
+    auto q = 0.5 * (-b - sgn(b) * std::sqrt(discriminant));
```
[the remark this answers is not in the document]
aprokop: The test passed on all backends. Wonder why that is.

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,300 @@
...
+
+  KOKKOS_FUNCTION static std::size_t
+  size(const Predicates<MemorySpace> &predicates)
+  {
+    return predicates._rays.extent(0);
+  }
+  KOKKOS_FUNCTION static auto get(Predicates<MemorySpace> const &predicates,
+                                  std::size_t i)
+  {
+    return attach(intersects(predicates._rays(i)), (int)i);
```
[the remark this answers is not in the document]
aprokop: Our callback interface is based on `size_t`, but we don't need to attach full 8 bytes, usually 4 for `int` is sufficient. We do this pretty commonly, for example, in DBSCAN.

examples/raytracing/example_raytracing.cpp
aprokop: No, removing this is wrong. Why would you want to do it?

masterleinad: I wanted to condense the initialization, see https://github.com/arborx/ArborX/pull/461#discussion_r567255945.

aprokop: Ah, I see what you mean. I still don't think it should be done. Conceptually, tmin and tmax are assigned to be the min and max roots of the quadratic equation. The logic dealing with uni-directionality of the ray should be separate from that.

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,300 @@
...
+  // If a solution does not exist, returns false.
+  KOKKOS_FUNCTION bool solveQuadratic(float a, float b, float c, float &x1,
+                                      float &x2) const
+  {
+    assert(a != 0);
+
+    auto discriminant = b * b - 4 * a * c;
+    if (discriminant < 0)
+      return false;
+    else if (discriminant == 0)
```
aprokop, line 96:
```suggestion
    if (discriminant == 0)
```

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,300 @@
...
+      x1 = x2 = -b / (2 * a);
+      return true;
+    }
+
+    // Instead of doing a simple
+    //   0.5 * (-b +- std::sqrt(discriminant)) / (2*a)
+    // we use a more stable algorithm with less loss of precision (see, for
+    // example, https://www.scratchapixel.com/lessons/3d-basic-rendering/
+    // minimal-ray-tracer-rendering-simple-shapes/ray-sphere-intersection).
+    auto sgn = [](float x) { return (x > 0) - (x < 0); };
```
aprokop, line 107:
```suggestion
    auto const sgn = [](float x) { return (int)((x > 0) - (x < 0)); };
```

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,300 @@
...
+{
+  using memory_space = MemorySpace;
+
+  KOKKOS_FUNCTION static std::size_t
+  size(const Predicates<MemorySpace> &predicates)
+  {
+    return predicates._rays.extent(0);
+  }
+  KOKKOS_FUNCTION static auto get(Predicates<MemorySpace> const &predicates,
+                                  std::size_t i)
```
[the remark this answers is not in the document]
aprokop: While you are right, and it should be done like that, I don't want to break off the style we used everywhere without `const`.

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,300 @@
...
+  // with
+  //     a2 = |d|^2, a1 = 2*(d, o - c), and a0 = |o - c|^2 - r^2.
+  // Then, we only need to intersect the solution interval [tmin, tmax] with
+  // [0, +inf) for the unidirectional ray.
+  KOKKOS_FUNCTION float overlapDistance(ArborX::Ray const &ray,
+                                        ArborX::Sphere const &sphere) const
+  {
+    auto const &c = sphere.centroid();
+    auto const &r = sphere.radius();
+    auto const &o = ray.origin();
```
Rombur, line 136: `o` is a bad name for a variable. Depending on the font it may look like `0`

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,300 @@
...
+  std::map<std::string, double> elapsed;
+
+  ExecutionSpace exec_space{};
+
+  exec_space.fence();
+  timer.reset();
+  ArborX::BVH<MemorySpace> bvh{exec_space, Primitives<MemorySpace>{spheres}};
+
+  Kokkos::View<float *, MemorySpace> accumulator("accumulator", num_rays);
+  bvh.query(exec_space, Predicates<MemorySpace>{rays},
```
Rombur, line 281: So what's our policy on `query`? @aprokop told me it was deprecated

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,300 @@
...
+    return v1[0] * v2[0] + v1[1] * v2[1] + v1[2] * v2[2];
+  }
+
+  // Solves a*x^2 + b*x + c = 0.
+  // If a solution exists, return true and stores roots at x1, x2.
+  // If a solution does not exist, returns false.
+  KOKKOS_FUNCTION bool solveQuadratic(float const a, float const b,
+                                      float const c, float &x1, float &x2) const
+  {
+    assert(a != 0);
```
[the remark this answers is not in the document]
aprokop: Hmm? We do that all the time in ArborX, don't we? See, for example, `src/details/ArborX_DetailsContainers.hpp`.

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,300 @@
...
+
+    auto const discriminant = b * b - 4 * a * c;
+    if (discriminant < 0)
+      return false;
+    if (discriminant == 0)
```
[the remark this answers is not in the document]
aprokop: Pretty low, unless it's `x^2 = 0`, which could happen.

examples/raytracing/example_raytracing.cpp
aprokop: No, we "deprecate" (or rather, discourage) queries that return CRS graph. The preferred way is to use a query that only takes in predicates and a callback, which is the case here.

Having said that, @dalg24's thinking is that we'll need to keep the version with CRS graph for a long time for backwards compatibility. I don't fully agree, but because of that it hasn't really been deprecated.

examples/raytracing/example_raytracing.cpp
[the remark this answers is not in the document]
masterleinad: Yes, but it's causing us issues with every new backend and it's one of the reasons that we need to define `DNDEBUG` everywhere.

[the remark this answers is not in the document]
Rombur: Well that's a bad idea. `assert` is not used in Kokkos except for HPX (which is host only) and SYCL.

examples/raytracing/example_raytracing.cpp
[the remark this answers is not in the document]
masterleinad: `(2ax+b)^2 == 0` :slightly_smiling_face:

examples/raytracing/example_raytracing.cpp
Rombur: > No, we "deprecate" (or rather, discourage) queries that return CRS graph. The preferred way is to use a query that only takes in predicates and a callback, which is the case here.

OK then we need to update the [wiki](https://github.com/arborx/ArborX/wiki/ArborX%3A%3AAccessTraits#example) because otherwise I don't see how we can ever remove that function.

examples/raytracing/example_raytracing.cpp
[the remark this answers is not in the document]
aprokop: If we replace it by `if(...) Kokkos::abort()`, we would need to protect each instance by `DNDEBUG`, if we don't want them in `Release` mode, right? We would likely need to introduce some macro for it, I suppose. Please make a decision together @dalg24 of how to proceed, I have no opinion on the matter.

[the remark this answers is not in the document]
masterleinad: (It's also not used within `SYCL` device code.)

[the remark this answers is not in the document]
masterleinad: See https://github.com/arborx/ArborX/pull/449.

examples/raytracing/example_raytracing.cpp
aprokop: See comment below.

masterleinad: I can live with either.

examples/raytracing/example_raytracing.cpp
aprokop: We have `ArborX::query` that takes in a spatial index structure to perform the same functionality. So we could, indeed, deprecate and remove it, if desired.

examples/raytracing/example_raytracing.cpp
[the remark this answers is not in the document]
aprokop: While you bring up valid points, this issue is beyond the scope of this PR, and should be addressed separately in the whole ArborX. So I would say we should ignore it here.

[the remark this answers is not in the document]
dalg24: Beyond the scope of this PR.  Let's accept the code as is and fix elsewhere.

examples/raytracing/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+add_executable(ArborX_RayTracing.exe example_raytracing.cpp)
+target_link_libraries(ArborX_RayTracing.exe ArborX::ArborX Boost::program_options)
+add_test(NAME ArborX_RayTracing_Example COMMAND ./ArborX_RayTracing.exe --spheres=100 --rays=10000 --L=100.0)
```
dalg24, line 3: Why are we passing command line arguments if we just specifies the default values?

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,299 @@
...
+  auto time = timer.seconds();
+
+  float accumulator_avg = 0.f;
+  Kokkos::parallel_reduce(
+      "ray_avg",
+      Kokkos::RangePolicy<ExecutionSpace>(exec_space, 0, accumulator.extent(0)),
+      KOKKOS_LAMBDA(int const i, float &update) {
+        update += accumulator(i) / num_rays;
+      },
+      accumulator_avg);
```
dalg24, line 292: Any reason to prefer this over `ArborX::accumulate(exec_space, accumulator, 0.f) / num_rays`?

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,299 @@
...
+    rays_host(i) = {
+        ArborX::Point{rand_uniform() * L, rand_uniform() * L, 0.f},
+        ArborX::Ray::Vector{float(std::cos(2 * M_PI * xi_2) * std::sqrt(xi_1)),
+                            float(std::sin(2 * M_PI * xi_2) * std::sqrt(xi_1)),
+                            std::sqrt(1.f - xi_1)}};
+  }
+  Kokkos::deep_copy(rays, rays_host);
+
+  Kokkos::Timer timer;
+  std::map<std::string, double> elapsed;
```
dalg24, line 271: Not used

dalg24 (commented):
Looks good overall.
I think it is unfortunate that you named your type `Primitives`, `Predicates`, and `Callback` (I would have tried something like `SpheresToBoxes`, `Rays`, and `AccumulateRaySphereIntersectionDistance`)

examples/raytracing/example_raytracing.cpp
aprokop: No, simply that I forgot that that thing exists.

examples/CMakeLists.txt
```diff
@@ -13,4 +13,5 @@ find_package(Boost COMPONENTS program_options)
 if(Boost_FOUND)
   add_subdirectory(viz)
   add_subdirectory(dbscan)
+  add_subdirectory(raytracing)
```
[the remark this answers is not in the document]
wjge: If there is no specific rule, I would keep it as raytracing, just like callback and dbscan.
The underscore can be saved for later when the example pool gets really big, so that you can do sth like dbscan_type1_variation3.

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,299 @@
...
+  size(const Primitives<MemorySpace> &primitives)
+  {
+    return primitives._spheres.extent(0);
+  }
+  KOKKOS_FUNCTION static ArborX::Box
+  get(Primitives<MemorySpace> const &primitives, std::size_t const i)
+  {
+    auto const &sphere = primitives._spheres(i);
+    auto const &c = sphere.centroid();
+    auto const &r = sphere.radius();
```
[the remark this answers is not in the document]
wjge: This has been updated.

examples/raytracing/example_raytracing.cpp
wjge: This has been updated.

examples/raytracing/example_raytracing.cpp
wjge: Why any of the compiler doesn't generate a warning to block this?

masterleinad:
The indentation if off.

examples/raytracing/example_raytracing.cpp
```diff
@@ -21,56 +21,56 @@
...
   }
   KOKKOS_FUNCTION static ArborX::Box
-  get(Primitives<MemorySpace> const &primitives, std::size_t const i)
+  get(SpheresToBoxes<MemorySpace> const &stobs, std::size_t const i)
   {
-    auto const &sphere = primitives._spheres(i);
-    auto const &c = sphere.centroid();
-    auto const &r = sphere.radius();
+    auto const &sphere = stobs._spheres(i);
+    auto const c = sphere.centroid();
```
aprokop, line 43: No, this was wrongly changed. Change back to `auto const &c = sphere.centroid();`, as it is a complex `Point` struct.

examples/raytracing/example_raytracing.cpp
aprokop: Not going to do this right now in this PR, can easily be added later.

aprokop:
@masterleinad @Rombur @dalg24 I think @wjge addressed all your comments. Please re-review this PR.

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,291 @@
...
+#include <ArborX.hpp>
+#include <ArborX_Version.hpp>
+
+#include <Kokkos_Core.hpp>
+
+#include <boost/program_options.hpp>
+
+#include <iostream>
+#include <random>
+#include <vector>
```
masterleinad, line 21: Where do you use `std::vector`?

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,291 @@
...
+  KOKKOS_FUNCTION float dotProduct(Vector const &v1, Vector const &v2) const
+  {
+    return v1[0] * v2[0] + v1[1] * v2[1] + v1[2] * v2[2];
+  }
+
+  // Solves a*x^2 + b*x + c = 0.
+  // If a solution exists, return true and stores roots at x1, x2.
+  // If a solution does not exist, returns false.
+  KOKKOS_FUNCTION bool solveQuadratic(float const a, float const b,
+                                      float const c, float &x1, float &x2) const
```
masterleinad, line 87:
```suggestion
  static KOKKOS_FUNCTION bool solveQuadratic(float const a, float const b,
                                      float const c, float &x1, float &x2)
```
?

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,291 @@
...
+  // Substituting (2) into (1),
+  //     |(o + t*d) - c|^2 - r^2 = 0,   (3)
+  // results in a quadratic equation for unknown t
+  //     a2 * t^2 + a1 * t + a0 = 0
+  // with
+  //     a2 = |d|^2, a1 = 2*(d, o - c), and a0 = |o - c|^2 - r^2.
+  // Then, we only need to intersect the solution interval [tmin, tmax] with
+  // [0, +inf) for the unidirectional ray.
+  KOKKOS_FUNCTION float overlapDistance(ArborX::Ray const &ray,
+                                        ArborX::Sphere const &sphere) const
```
masterleinad, line 130:
```suggestion
  static KOKKOS_FUNCTION float overlapDistance(ArborX::Ray const &ray,
                                        ArborX::Sphere const &sphere)
```
?

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,291 @@
...
+  }
+};
+
+template <typename MemorySpace>
+struct AccumRaySphereInterDist
+{
+  using Vector = ArborX::Ray::Vector;
+
+  Kokkos::View<ArborX::Sphere *, MemorySpace> _spheres;
+  Kokkos::View<float *, MemorySpace> _accumulator;
```
Rombur, line 76: Remove the `_` of these two variables

Rombur (changes requested):
The code looks fine except for the `_` which does not follow our code style https://github.com/arborx/ArborX/wiki/CodeStyle

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,291 @@
...
+#include <boost/program_options.hpp>
+
+#include <iostream>
+#include <random>
+#include <vector>
+
+template <typename MemorySpace>
+struct SpheresToBoxes
+{
+  Kokkos::View<ArborX::Sphere *, MemorySpace> _spheres;
```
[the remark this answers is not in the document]
aprokop: Do we? I don't think we have consistent style for that. For example, `ArborX::Box` has public `_min_corner` and `_max_corner`, and `ArborX::Sphere` has public `_centroid` and `_radius`. Personally, I was not aware of this differentiation. For me, the logic is that `_` means it is a struct/class member, does not matter public/private (though, we are not consistent in this sense either).

examples/raytracing/example_raytracing.cpp
aprokop: Don't think it matters. This function will only be called internally from a non-static one.

masterleinad: Otherwise, it wouldn't even compile. :wink:

examples/raytracing/example_raytracing.cpp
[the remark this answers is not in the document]
Rombur: > For example, ArborX::Box has public _min_corner and _max_corner, and ArborX::Sphere has public _centroid and _radius.

If I remember correctly, we wanted them to be private but I think that there was a problem with list initialization. Note that even in this example, you don't access `_centroid` and `_radius` directly but instead you use the function `centroid()` and `radius()`.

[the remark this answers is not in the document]
aprokop: You may be right, there was something like that (lambdas on Cuda, if I recall). I am still unsure we have a consistent style for this in ArborX, and this should not block the PR. I also personally am against the proposed style of naming public members without underscore, as this makes them indistinguishable from other variables.

dalg24:
Let's cleanup history and merge.  We can revisit the naming later.

src/details/ArborX_Ray.hpp
```diff
@@ -136,6 +148,96 @@ bool intersects(Ray const &ray, Box const &box)
...
+
+  // Instead of doing a simple
+  //    (-b +- std::sqrt(discriminant)) / (2*a)
+  // we use a more stable algorithm with less loss of precision (see, for
+  // example, https://www.scratchapixel.com/lessons/3d-basic-rendering/
+  // minimal-ray-tracer-rendering-simple-shapes/ray-sphere-intersection).
+  auto const sgn = [](float x) { return (int)(x > 0) - (int)(x < 0); };
+  auto const q = 0.5 * (-b - sgn(b) * std::sqrt(discriminant));
+  x1 = q / a;
+  x2 = c / q;
```
aprokop, line 182: @dalg24 discovered that this code contains a bug. Specifically, when `b = 0`, `q = inf`, and the roots are wrong. This case has to be treated specially.

aprokop (changes requested):
Need to fix the bug.

test/tstDetailsRay.cpp
```diff
@@ -158,3 +157,65 @@ BOOST_AUTO_TEST_CASE(intersects_box)
...
+  // behind the origin
+  BOOST_TEST(overlapDistance(Ray{{-2, 0, 0}, {-1, 0, 0}}, unit_sphere) == 0);
+  BOOST_TEST(overlapDistance(Ray{{0, 2, 2}, {0, 1, 1}}, unit_sphere) == 0);
+  BOOST_TEST(overlapDistance(Ray{{0, 2, 2}, {0, 1, 1}}, unit_sphere) == 0);
+
+  // origin inside
+  auto const inf = KokkosExt::ArithmeticTraits::infinity<float>::value;
+  BOOST_TEST(overlapDistance(Ray{{0.5, 0, 0}, {0, 1, 1}}, unit_sphere) == inf);
+  BOOST_TEST(overlapDistance(Ray{{0, 0.25, 0}, {0, 0, 1}}, unit_sphere) == inf);
+  BOOST_TEST(overlapDistance(Ray{{0, 0, 0}, {1, 2, 3}}, unit_sphere) == inf);
```
aprokop, line 195: Once the bug is fixed, these have to be changed.

test/tstDetailsRay.cpp
```diff
@@ -158,3 +157,65 @@ BOOST_AUTO_TEST_CASE(intersects_box)
...
+
+  // origin on surface
+  auto const half_sqrtf_2 = std::sqrt(2.f) / 2;
+  auto const half_sqrtf_3 = std::sqrt(3.f) / 2;
+  //   tangent
+  BOOST_TEST(overlapDistance(Ray{{0, 0, 1}, {1, -1, 0}}, unit_sphere) == 0);
+  BOOST_TEST(overlapDistance(Ray{{0, 1, 0}, {2, 0, 3}}, unit_sphere) == 0);
+  BOOST_TEST(overlapDistance(Ray{{half_sqrtf_2, half_sqrtf_2, 0}, {1, -1, 0}},
+                             unit_sphere) ==
+             inf); // inside at machine precision...
```
aprokop, line 205: This one may need to be fixed.

src/details/ArborX_Ray.hpp
wjge: b=0 alone is not a problem, maybe sgn(b) returning 0 is an issue. We can use the the below implementation instead:
 q = (b > 0) ? 
            -0.5 * (b + sqrt(discriminant)) : 
            -0.5 * (b - sqrt(discriminant));

aprokop: Yeah, I think this could work, too. You had it originally, and I tried to be cleaner and introduced a bug.

aprokop: I would change it to
```c++
  q = (b > 0) ?  (-b - sqrt(discriminant))/2 : (-b + sqrt(discriminant))/2
```
to make it more in line with the standard formula `(-b +- sqrt(discriminant))/(2*a)`

examples/raytracing/example_raytracing.cpp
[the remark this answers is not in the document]
wjge: Shall we postpone this naming convention issue to a future PR?

masterleinad:
```
/var/jenkins/workspace/ArborX_PR-461/test/tstDetailsRay.cpp(172): error: in "overlap_distance_sphere": check overlapDistance(Ray{{-2, -2, -2}, {1, 1, 1}}, unit_sphere) == 2 has failed [2.00000024 != 2]
/var/jenkins/workspace/ArborX_PR-461/test/tstDetailsRay.cpp(209): error: in "overlap_distance_sphere": check overlapDistance(Ray{{half_sqrtf_2, half_sqrtf_2, 0}, {1, -1, 0}}, unit_sphere) == 0 has failed [0.000244140625 != 0]
/var/jenkins/workspace/ArborX_PR-461/test/tstDetailsRay.cpp(218): error: in "overlap_distance_sphere": check overlapDistance(Ray{{0, -half_sqrtf_2, half_sqrtf_2}, {0, 1, -1}}, unit_sphere) == 2 has failed [1.99999988 != 2]
```

aprokop (commented):
I'm fine with the updated version (moving `Ray` to `Experimental`). The outstanding issue is specifying tolerance. @dalg24 is on the case.

test/tstDetailsRay.cpp
```diff
@@ -158,3 +157,74 @@ BOOST_AUTO_TEST_CASE(intersects_box)
   BOOST_TEST(!intersects(Ray{{1, 1.5, 1}, {0, -1, -2.1}}, unit_box));
   BOOST_TEST(!intersects(Ray{{1, 1.5, 1}, {-2.1, -1, 0}}, unit_box));
 }
+
+// NOTE until boost 1.70 need to cast both operands when comparing floating
+// points
+BOOST_AUTO_TEST_CASE(overlap_distance_sphere,
+                     *boost::unit_test::tolerance(
+                         boost::test_tools::fpc::percent_tolerance(0.0001f)))
```
aprokop, line 165:
```suggestion
                     *boost::unit_test::tolerance(1e-6f))

```
From [here](https://www.boost.org/doc/libs/1_68_0/libs/test/doc/html/boost_test/testing_tools/extended_comparison/floating_point.html), these are equivalent. I also tested it locally.

test/tstDetailsRay.cpp
```diff
@@ -158,3 +157,74 @@ BOOST_AUTO_TEST_CASE(intersects_box)
...
+  BOOST_TEST(overlapDistance(Ray{{0, 0, 0}, {1, 2, 3}}, unit_sphere) == 1.f);
+
+  // origin on surface
+  //   tangent
+  BOOST_TEST(overlapDistance(Ray{{0, 0, 1}, {1, -1, 0}}, unit_sphere) == 0.f);
+  BOOST_TEST(overlapDistance(Ray{{0, 1, 0}, {2, 0, 3}}, unit_sphere) == 0.f);
+  BOOST_TEST(
+      overlapDistance(Ray{{half_sqrtf_2, half_sqrtf_2, 0}, {1, -1, 0}},
+                      unit_sphere) == 0.f,
+      boost::test_tools::tolerance(0.0005f)); // inside at machine precision...
```
aprokop, line 214:
```suggestion
      boost::test_tools::tolerance(5e-4f)); // inside at machine precision...
```

aprokop (approved):

examples/raytracing/example_raytracing.cpp
```diff
@@ -0,0 +1,200 @@
...
+      Kokkos::ViewAllocateWithoutInitializing("rays"), num_rays);
+  auto rays_host = Kokkos::create_mirror_view(rays);
+
+  for (int i = 0; i < num_rays; ++i)
+  {
+    float xi_1 = rand_uniform();
+    float xi_2 = rand_uniform();
+
+    rays_host(i) = {ArborX::Point{rand_uniform() * L, rand_uniform() * L, 0.f},
+                    ArborX::Experimental::Ray::Vector{
```
dalg24, line 170: Not asking to change but mentioning we do not want to expose `Ray::Vector`

aprokop: That is correct. `Vector` is really unfinished business right now, and more of a hack than a proper concept.

PR #465, opened 2021-01-21
Breakout KokkosExt header and partial cleanup of includes

dalg24:
Motivation: While profiling the compilation of some of our executables, I realized `<ArborX_Box.hpp>` was including the whole `<Kokkos_Core.hpp>` which is really silly. This PR is not attempting to fix all the defects, it specifically resolves that issue by breaking out the header `<ArborX_DetailsKokkosExt.hpp>` into
* `ArborX_DetailsKokkosExtAccessibilityTraits.hpp`
* `ArborX_DetailsKokkosExtArithmeticTraits.hpp` (which eventually will be replaced by Kokkos numeric traits)
* `ArborX_DetailsKokkosExtMathFunctions.hpp` (which eventually will be replaced by the math functions in `Kokkos::`)
* `ArborX_DetailsKokkosExtMinMaxOperations.hpp` (replaced by `std:{min,max}` if NVCC quits requiring an experimental flag to enable)

### DIFF
--- a/src/ArborX_LinearBVH.hpp
+++ b/src/ArborX_LinearBVH.hpp
@@ -18,7 +18,7 @@
 #include <ArborX_CrsGraphWrapper.hpp>
 #include <ArborX_DetailsBatchedQueries.hpp>
 #include <ArborX_DetailsConcepts.hpp>
-#include <ArborX_DetailsKokkosExt.hpp>
+#include <ArborX_DetailsKokkosExtAccessibilityTraits.hpp>
 #include <ArborX_DetailsNode.hpp>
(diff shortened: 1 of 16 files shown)


aprokop (commented):
Don't know what to think of this. Looks like noise to me. Moving arithmetic traits is a bit more palatable.

Is this being done to accelerate some builds in the unit testing? For an ArborX user, I don't see anything changing.

dalg24:
> Is this being done to accelerate some builds in the unit testing? 

Yes

> For an ArborX user, I don't see anything changing.

If they only need `Box` in a translation unit they will certainly see the difference.

aprokop (approved):

PR #466, opened 2021-01-21
Move comparison function objects out of priority queue header

dalg24:
Rational: It somehow bothered me when we had to include `<ArborX_DetailsPriorityQueue.hpp>` to use `Less` in #453 

Note that I made the operators `constexpr` to be consistent with the function objects in `<functional>`

### DIFF
--- a/examples/dbscan/dbscan.cpp
+++ b/examples/dbscan/dbscan.cpp
@@ -11,7 +11,7 @@
 
 #include <ArborX_DBSCAN.hpp>
 #include <ArborX_DetailsHeap.hpp>
-#include <ArborX_DetailsPriorityQueue.hpp> // Less
+#include <ArborX_DetailsOperatorFunctionObjects.hpp> // Less
 #include <ArborX_Version.hpp>
 
 #include <Kokkos_Core.hpp>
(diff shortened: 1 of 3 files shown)


aprokop (commented):
I agree with the idea. I'm unused to "function object", and not sure what it is. If it is not commonly used, I would prefer something else like `ArborX_DetailsComparisonFunctors.hpp`.

And do you envision the scope of this file to expand in the future to include something else?

dalg24:
> I agree with the idea. I'm unused to "function object", and not sure what it is. If it is not commonly used, I would prefer something else like `ArborX_DetailsComparisonFunctors.hpp`.

https://eel.is/c++draft/function.objects

> And do you envision the scope of this file to expand in the future to include something else?

Scope no.  More function objects maybe.

I should have commented that if it wasn't for NVCC we would use them directly from the standard.
https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#options-for-altering-compiler-linker-behavior-expt-relaxed-constexpr

dalg24:
> > I agree with the idea. I'm unused to "function object", and not sure what it is. If it is not commonly used, I would prefer something else like `ArborX_DetailsComparisonFunctors.hpp`.
> 
> https://eel.is/c++draft/function.objects
> 
> > And do you envision the scope of this file to expand in the future to include something else?
> 
> Scope no. More function objects maybe.

If you feel strongly about it I can change.  The name your suggest works.  I went for something more generic because I felt if we were to need `std::plus` I would not want to have it in a separate header.

aprokop (approved):

PR #467, opened 2021-01-21
Drop unused type traits

dalg24:
Rational: The traits have been unused for a long time and I don't anticipate we will change our mind anytime soon.

* Besides not being used, `is_expandable` and `has_centroid` functionality can easily be implemented using the detection idiom.
* `is_complete` had other issues that led to our using another mechanism to detect whether the access traits were specialized. https://github.com/arborx/ArborX/blob/b020ca71f3994d9bc8c0c3b151dceb41114061e3/src/details/ArborX_AccessTraits.hpp#L35

### DIFF
--- a/src/details/ArborX_AccessTraits.hpp
+++ b/src/details/ArborX_AccessTraits.hpp
@@ -8,10 +8,11 @@
  *                                                                          *
  * SPDX-License-Identifier: BSD-3-Clause                                    *
  ****************************************************************************/
+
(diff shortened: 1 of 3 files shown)


dalg24:
The CUDA-10.2-NVCC failure was fetnat03 running out of space

aprokop (approved):
Honestly, did not even know they existed 🤷

PR #473, opened 2021-02-03
Avoid unnecessary rebuilds for generated unit test source files

dalg24:
Note that CMake 3.18 introduces
```cmake
file(CONFIGURE OUTPUT <output-file> CONTENT <content> [...])
```


aprokop (approved):
Per [CMake](https://cmake.org/cmake/help/latest/command/file.html#write):
```
file(WRITE <filename> <content>...)

Write <content> into a file called <filename>. <snip>

If the file is a build input, use the configure_file() command to update the file only when its content changes.
```

PR #479, opened 2021-02-16
Avoid cluttering the console log in annotations tests

dalg24:
Use `BOOST_TEST_MESSAGE` instead of printing to the standard outpout.
One can pass `--log_level=message` to get the verbose output.


masterleinad (approved):
Looks good to me.

PR #484, opened 2021-02-25
Add __pycache__ git gitignore

aprokop:


dalg24 (commented):
This is something you can setup locally, whether you want it systemwide or just for a specific repository.  See the [documentation](https://git-scm.com/docs/gitignore). If it was up to me I would remove that file altogether.

aprokop:
I find it convenient, and would like to have it merged.

PR #486, opened 2021-03-09
Require CMake 3.16

dalg24:
Per our discussion at the developers meeting.
Kokkos also [requires 3.16 as the minimum version](https://github.com/kokkos/kokkos/blob/897b79da51c1719b620990de4eabba154094bffe/CMakeLists.txt#L75) and I would like several features that are only available from that version (target-dependent generator expression in `add_test`, support for faster builds, etc.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.12)
+cmake_minimum_required(VERSION 3.16)
 project(ArborX CXX)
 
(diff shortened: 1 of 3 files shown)


aprokop (approved):
I don't have a strong opinion, and am fine with this change.

Rombur (approved):

Rombur:
Rebuild this please

masterleinad (approved):

PR #498, opened 2021-03-25
Avoid printing to the standard output in the tests

dalg24:


test/tstDetailsTreeConstruction.cpp
```diff
@@ -18,16 +18,13 @@
...
-#include <algorithm>
 #include <array>
 #include <bitset>
 #include <functional>
 #include <limits>
 #include <sstream>
 #include <vector>
 
-#define BOOST_TEST_MODULE DetailsTreeConstruction
-
```
masterleinad, line 30: Does this change belong here?

masterleinad (commented):
Apart from the one question, this looks good.

test/tstDetailsTreeConstruction.cpp
dalg24: This is arguable.  I am editing that file so I figure I might as well fix it.
I can do it elsewhere if you want me to.

masterleinad (approved):
Looks OK to me.

aprokop (commented):
What are the advantages of using `BOOST_MESSAGE` over `std::cout`?

Rombur:
You can decide at runtime if the message should be printed on the screen

Rombur (approved):

dalg24:
Adding reference to #479

PR #499, opened 2021-04-05
Update CI to ROCm 4.2

Rombur:
ROCm 4.1 fixes a lot of register of spills. Let's see if that fixes the CI

### DIFF
--- a/.jenkins
+++ b/.jenkins
@@ -327,12 +327,12 @@ pipeline {
                     }
                 }
 
-                stage('HIP-4.0') {
+                stage('HIP-4.2') {
                     agent {
                         dockerfile {
                             filename "Dockerfile.hipcc"
                             dir "docker"
-                            additionalBuildArgs '--build-arg BASE=rocm/dev-ubuntu-20.04:4.0 --build-arg KOKKOS_ARCH=${KOKKOS_ARCH}'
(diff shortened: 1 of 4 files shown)


dalg24 (approved):

dalg24:
Retest this please

masterleinad:
Apparently, the `HIP` build is still stuck. :slightly_frowning_face:

Rombur:
It might be because of the OS. ROCm requires a newer version RHEL (at least 8.2). lascaux has centos 7.9 I wonder if that has an impact of the drivers.

Rombur:
I found this in the [release note](https://github.com/RadeonOpenCompute/ROCm#containers):

```
A container (built with error detection for this issue) using a ROCm v4.1 or newer run-time is initiated to execute on an older kernel. The container fails to start and the following warning appears:

Error: Incompatible ROCm environment. The Docker container requires the latest kernel driver to operate correctly. Upgrade the ROCm kernel to v4.1 or newer, or use a container tagged for v4.0.1 or older.
```

Looks like we need to update drivers.

Rombur:
This update the CI to rocm 4.2. The version of the compiler should be bug free so if there are problems with the hip backend it's in our side.

aprokop (approved):
Looks good. Please update the wiki's Build page.

PR #516, opened 2021-05-06
Fixup Kokkos space accessibility checks

dalg24:
Rational: `Kokkos::Impl:: SpaceAccessibility ` is being marked as deprecated.  One is suppose to use `Kokkos:: SpaceAccessibility` instead (https://github.com/kokkos/kokkos/pull/3978).  As I was making sure we are not using it, I discovered we were using `Kokkos::Impl::MemorySpaceAccess` and we should not do that.


aprokop (approved):
LGTM. Please squash commits 1, 2, and 4.

dalg24:
> Please squash commits 1, 2, and 4.

1, 4 do the same thing.  2 is different though.  You can squash if you like.

masterleinad (approved):

PR #517, opened 2021-05-06
[WIP] Ray-triangle intersections

dalg24:
@wjge please review

### DIFF
--- a/src/details/ArborX_Ray.hpp
+++ b/src/details/ArborX_Ray.hpp
@@ -275,42 +275,52 @@ KOKKOS_INLINE_FUNCTION float overlapDistance(Ray const &ray,
 }
 
 // Möller–Trumbore intersection algorithm
-KOKKOS_INLINE_FUNCTION bool intersects(Ray const &ray, Triangle const &triangle)
+KOKKOS_INLINE_FUNCTION bool rayTriangleIntersect(Ray const &ray,
+                                                 Triangle const &triangle,
+                                                 float &t, float &u, float &v)
 {
   auto ab = makeVector(triangle.a, triangle.b);
   auto ac = makeVector(triangle.a, triangle.c);
 
-  auto h = crossProduct(ray.direction(), ac);
-  auto a = dotProduct(ab, h);
+  auto p = crossProduct(ray.direction(), ac);
+  auto det = dotProduct(ab, p);
 
   auto const epsilon = 0.0000001f;
-  if (a > -epsilon && a < epsilon)
+  // If the determinant is negative the triangle is back-facing.
+  // If the determinant is close to 0, the ray misses the triangle because they
+  // are in the same plane.
+  if (det < epsilon)
   {
-    return false; // ray parallel to the triangle
+    return false;
   }
 
-  auto f = 1.f / a;
+  float const inv_det = 1.f / det;
+
   auto s = makeVector(triangle.a, ray.origin());
-  auto u = f * dotProduct(s, h);
+  u = inv_det * dotProduct(s, p);
   if (u < 0.f || u > 1.f)
   {
     return false;
   }
 
   auto q = crossProduct(s, ab);
-  auto v = f * dotProduct(ray.direction(), q);
+  v = inv_det * dotProduct(ray.direction(), q);
   if (v < 0.f || u + v > 1.f)
   {
     return false;
   }
 
-  auto t = f * dotProduct(ac, q);
-  if (t > epsilon)
-  {
-    return true;
-  }
+  t = inv_det * dotProduct(ac, q);
 
-  return false;
+  return true;
+}
+
+KOKKOS_INLINE_FUNCTION bool intersects(Ray const &ray, Triangle const &triangle)
+{
+  float t;
+  float u;
+  float v;
+  return rayTriangleIntersect(ray, triangle, t, u, v);
 }
 
 } // namespace Experimental
--- a/src/details/ArborX_Ray.hpp
+++ b/src/details/ArborX_Ray.hpp
@@ -279,31 +279,31 @@ KOKKOS_INLINE_FUNCTION bool rayTriangleIntersect(Ray const &ray,
                                                  Triangle const &triangle,
                                                  float &t, float &u, float &v)
 {
-  auto ab = makeVector(triangle.a, triangle.b);
-  auto ac = makeVector(triangle.a, triangle.c);
+  auto const ab = makeVector(triangle.a, triangle.b);
+  auto const ac = makeVector(triangle.a, triangle.c);
 
-  auto p = crossProduct(ray.direction(), ac);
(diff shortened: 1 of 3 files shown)


src/details/ArborX_Ray.hpp
```diff
@@ -273,6 +274,45 @@ KOKKOS_INLINE_FUNCTION float overlapDistance(Ray const &ray,
   return (tmax - tmin);
 }
 
+// Möller–Trumbore intersection algorithm
```
Rombur, line 277: I found a pretty nice explanation of the algorithm [here](https://www.scratchapixel.com/lessons/3d-basic-rendering/ray-tracing-rendering-a-triangle/moller-trumbore-ray-triangle-intersection). If you scroll down, they have their source code which uses better names for the variable.

src/details/ArborX_Ray.hpp
```diff
@@ -273,6 +274,45 @@ KOKKOS_INLINE_FUNCTION float overlapDistance(Ray const &ray,
...
+  }
+
+  auto q = crossProduct(s, ab);
+  auto v = f * dotProduct(ray.direction(), q);
+  if (v < 0.f || u + v > 1.f)
+  {
+    return false;
+  }
+
+  auto t = f * dotProduct(ac, q);
```
Rombur, line 307: Is there a way to have access to this variable? It's the distance between the origin point and the intersection with the triangle. So we can simply get the intersection point using `origin + t * ray.direction()`.

dalg24: What would you want to set it to for any of the early returns above?

Rombur: I guess it would make sense to set that to infty since the line never intersects.

wjge (commented):
How do you usually pick a, b, c? It looks random to me by input. Do we need any rule at this moment?

dalg24:
> How do you usually pick a, b, c? It looks random to me by input. Do we need any rule at this moment?

I am not sure I understand your question.  Triangles are user input.

src/details/ArborX_Ray.hpp
```diff
@@ -273,6 +274,55 @@ KOKKOS_INLINE_FUNCTION float overlapDistance(Ray const &ray,
   return (tmax - tmin);
 }
 
+// Möller–Trumbore intersection algorithm
+KOKKOS_INLINE_FUNCTION bool rayTriangleIntersect(Ray const &ray,
```
dalg24, line 278: Note that this could/should be `constexpr`

src/details/ArborX_Ray.hpp
```diff
@@ -273,6 +274,55 @@ KOKKOS_INLINE_FUNCTION float overlapDistance(Ray const &ray,
...
+  auto ac = makeVector(triangle.a, triangle.c);
+
+  auto p = crossProduct(ray.direction(), ac);
+  auto det = dotProduct(ab, p);
+
+  auto const epsilon = 0.0000001f;
+  // If the determinant is negative the triangle is back-facing.
+  // If the determinant is close to 0, the ray misses the triangle because they
+  // are in the same plane.
+  if (det < epsilon)
```
Rombur, line 292: I think I prefer the previous condition. This could lead to unexpected behavior for users.

dalg24: Fixed

dalg24:
> In the M-T algorithm in this PR, only the edges ab and ac are used. It is possible to describe a triangle with point a, vector ab, vector ac

This is a good point.  Are they other algorithms we might want to support for triangles?  We could also merge in the current version and improve later when we measure performance.  I am actually curious how much this matters.

> Triangle class as a base class 

Base class for what?  Keep in mind virtual inheritance can be problematic on the device.

> does ab cross ac need to be positive by default for double-face triangle?

Have not put much thought into this

> 2-D geometry

I would not worry about it for now.  We would probably call it something else in 2D anyway.

Rombur:
Please retest this.

Rombur:
The tests pass locally for me :/

src/details/ArborX_Ray.hpp
```diff
@@ -273,6 +274,58 @@ KOKKOS_INLINE_FUNCTION float overlapDistance(Ray const &ray,
...
+  auto const ac = makeVector(triangle.a, triangle.c);
+
+  auto const p = crossProduct(ray.direction(), ac);
+  auto const det = dotProduct(ab, p);
+
+  auto const epsilon = 0.0000001f;
+  // If the determinant is negative the triangle is back-facing.
+  // If the determinant is close to 0, the ray misses the triangle because they
+  // are in the same plane.
+  if (det > -epsilon && det < epsilon)
```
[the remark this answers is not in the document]
wjge: Can we add more tests for this? I am thinking about have a series of similar triangles but in different scales.

Rombur:
I found the problem and it's not in this PR. The problem is when building a `Ray` we normalize the direction, i.e. we divide by the square root of the norm, using `float`. This introduces a small error in the direction of the `Ray` which then misses the triangle. Computing the square root using a `double` and then casting it to a `float` is not good enough. The entire normalization needs to be done using `double`. The rest of the computation can still be done using `float`. Since the problem is not from this PR itself, I'll create a new PR that fixes the problem.

src/details/ArborX_Ray.hpp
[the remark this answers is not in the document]
Rombur: There is a problem with large and small triangles but the problem is not due to the `epsilon`. The problem is that we are using `float` and not `double` and so we accumulate round-off errors. The only solution is to use `double` everywhere. If we mix `float` and `double` some of the errors does not cancel each other and the result is worse. This means that `Vector` needs to use `double` instead of `float`. I don't know if it's something we want to do though.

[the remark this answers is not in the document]
wjge: Can we print some warnings under certain conditions when the algorithm with float tends to have more misjudgment?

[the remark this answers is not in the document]
Rombur: It seems that the error in the dot product is a good indicator that there is a potential problem so we could do it. However, that means that we need to compare the dot product using `float` with the one using `double` so it's more costly. Also, there are cases where there is an error on the dot product but it doesn't change the final result.

Rombur:
@dalg24 can you rebase this on master or do you want me to do it? That should fix the CI and then, we should be able to merge this.

Rombur (commented):
I have started using the PR with real data and it doesn't work. The problem is that if the ray is in the same plane as the triangle, the algorithm assumes that the ray and the triangle do not intersect. We need to add a check when `det` is small that the ray really misses the triangle.

aprokop:
Should this PR be closed given we have two superseding PRs (#604, #617)?

Rombur:
Yes, we can close it. #604 is based on this PR.

PR #520, opened 2021-05-11
Restrict HIP builds to MI60

masterleinad:
In all releases, `Kokkos` requires to set a `HIP` architecture but our CI for HIP doesn't have a uniform architecture. At the moment we specify, that we build for MI60 so we should also make sure that the CI only runs on the machine that has a MI60.


masterleinad:
Yes, see https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/ArborX/detail/PR-386/46/tests.

masterleinad:
> The MI60 are going down this week (probably) for an unknown period of time (could be a few hours or a few days)

We can of course just test MI100 instead.

masterleinad:
The last commit only fails due to a warning when building Boost. [`f2ec019` (#386)](https://github.com/arborx/ArborX/pull/386/commits/f2ec019183a010daaee991fd7627c31a488df035) is the alternative using MI100. Of course, we should decide on this one first but I wanted to show both options.

PR #562, opened 2021-10-20
Cleanup box-ray intersection

dalg24:
Minor (hopefully uncontroversial) code cleanup before upcoming PR

The only change worth mentioning IMO is not bothering with `isnan` since the comparison would evaluate to false and the assignment would not happen anyway.


wjge:
I don't remember exactly the conditions. I remember I pushed the PGI compiler flag to make it pass the CI but I don't have the permission. Need to check the previous PR to find out.

PR #552, opened 2021-10-01
Index-based traversal

dalg24:
If not slower than current implementation, this makes it easier to explore different data structures for the BVH

PR #563, opened 2021-10-21
Check valid callback and space accessibility

dalg24:
I realized `check_valid_callback()` wasn't called when I forgot the `const`-qualifier on the `operator()` of a callback.  I also added a few static assertions to check space accessibility.

Note the last commit, removing the (unused) private member function `BVH::getRoot()`, oversight from #552.  Let me know if you want me to open a separate PR for it.


dalg24:
Failure was unrelated ("No space left on device" on fetnat03)

PR #572, opened 2021-11-05
Cleanup `Experimental::Ray`

dalg24:
* Drop backward compatibility `Ray::Vector` member type
* Let `Ray::{norm,normalize}` member functions be private


wjge:
Do we need to change std::sqrt to the Kokkos one?

dalg24:
> Do we need to change std::sqrt to the Kokkos one?

Eventually yes.  The issue is that it was added in Kokkos v3.4 and ArborX only requires a minimum version of 3.1 at the moment.

aprokop (approved):
LGTM

PR #600, opened 2021-12-22
Add functionality to compute mutual reachability distances

dalg24:
Subset of #557

### DIFF
--- a/src/details/ArborX_DetailsMutualReachabilityDistance.hpp
+++ b/src/details/ArborX_DetailsMutualReachabilityDistance.hpp
@@ -49,7 +49,7 @@ template <class Primitives>
 struct NearestK
 {
   Primitives primitives;
-  int k;
+  int k; // including self-collisions
 };
 
 } // namespace Details
@@ -67,8 +67,7 @@ struct AccessTraits<Details::NearestK<Primitives>, PredicatesTag>
   static KOKKOS_FUNCTION auto get(Details::NearestK<Primitives> const &x,
                                   size_type i)
   {
-    // NOTE account for self collision
-    return attach(nearest(Access::get(x.primitives, i), x.k + 1), i);
+    return attach(nearest(Access::get(x.primitives, i), x.k), i);
   }
 };
 
--- a/src/details/ArborX_DetailsMutualReachabilityDistance.hpp
+++ b/src/details/ArborX_DetailsMutualReachabilityDistance.hpp
@@ -38,7 +38,10 @@ struct MaxDistance
     using KokkosExt::max;
     auto const distance_ij =
         distance(Access::get(_primitives, i), Access::get(_primitives, j));
-    _distances(j) = max(_distances(j), distance_ij); // careful here
+    // NOTE using knowledge that each nearest predicate traversal is performed
+    // by a single thread.  The distance update below would need to be atomic
+    // otherwise.
+    _distances(j) = max(_distances(j), distance_ij);
   }
 };
 
--- a/src/details/ArborX_DetailsMutualReachabilityDistance.hpp
+++ b/src/details/ArborX_DetailsMutualReachabilityDistance.hpp
@@ -1,0 +1,93 @@
+/****************************************************************************
+ * Copyright (c) 2017-2021 by the ArborX authors                            *
+ * All rights reserved.                                                     *
+ *                                                                          *
+ * This file is part of the ArborX library. ArborX is                       *
+ * distributed under a BSD 3-clause license. For the licensing terms see    *
+ * the LICENSE file in the top-level directory.                             *
+ *                                                                          *
+ * SPDX-License-Identifier: BSD-3-Clause                                    *
+ ****************************************************************************/
+
+#ifndef ARBORX_DETAILS_MUTUAL_REACHABILITY_DISTANCE_HPP
+#define ARBORX_DETAILS_MUTUAL_REACHABILITY_DISTANCE_HPP
+
+#include <ArborX_AccessTraits.hpp>
+#include <ArborX_DetailsAlgorithms.hpp>
+#include <ArborX_DetailsHappyTreeFriends.hpp>
+#include <ArborX_DetailsKokkosExtMinMaxOperations.hpp>
+#include <ArborX_Predicates.hpp>
+
+namespace ArborX
+{
+namespace Details
+{
+
+template <class Primitives, class Distances>
+struct MaxDistance
+{
+  Primitives _primitives;
+  Distances _distances;
+  using Access = AccessTraits<Primitives, PrimitivesTag>;
+  using memory_space = typename Access::memory_space;
+  using size_type = typename memory_space::size_type;
+  template <class Predicate>
+  KOKKOS_FUNCTION void operator()(Predicate const &predicate, size_type i) const
+  {
+    size_type const j = getData(predicate);
+    using KokkosExt::max;
+    auto const distance_ij =
+        distance(Access::get(_primitives, i), Access::get(_primitives, j));
+    _distances(j) = max(_distances(j), distance_ij); // careful here
+  }
+};
+
+template <class Primitives>
+struct NearestK
+{
+  Primitives primitives;
+  int k;
+};
+
+} // namespace Details
+
+template <class Primitives>
+struct AccessTraits<Details::NearestK<Primitives>, PredicatesTag>
+{
+  using Access = AccessTraits<Primitives, PrimitivesTag>;
+  using memory_space = typename Access::memory_space;
+  using size_type = typename memory_space::size_type;
+  static KOKKOS_FUNCTION size_type size(Details::NearestK<Primitives> const &x)
+  {
+    return Access::size(x.primitives);
+  }
+  static KOKKOS_FUNCTION auto get(Details::NearestK<Primitives> const &x,
+                                  size_type i)
+  {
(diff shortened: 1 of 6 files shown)


src/details/ArborX_DetailsMutualReachabilityDistance.hpp
```diff
@@ -0,0 +1,93 @@
...
+  using size_type = typename memory_space::size_type;
+  static KOKKOS_FUNCTION size_type size(Details::NearestK<Primitives> const &x)
+  {
+    return Access::size(x.primitives);
+  }
+  static KOKKOS_FUNCTION auto get(Details::NearestK<Primitives> const &x,
+                                  size_type i)
+  {
+    // NOTE account for self collision
+    return attach(nearest(Access::get(x.primitives, i), x.k + 1), i);
```
aprokop, line 68:
```suggestion
    return attach(nearest(Access::get(x.primitives, i), x.k), i);
```
Core distance definition includes self collisions.

test/CMakeLists.txt
```diff
@@ -161,7 +161,11 @@ target_compile_definitions(ArborX_Clustering.exe PRIVATE BOOST_TEST_DYN_LINK)
 target_include_directories(ArborX_Clustering.exe PRIVATE ${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_SOURCE_DIR}/examples/dbscan)
 add_test(NAME ArborX_Clustering_Test COMMAND ./ArborX_Clustering.exe)
 
-add_executable(ArborX_DetailsTreeNodeLabeling.exe tstDetailsTreeNodeLabeling.cpp utf_main.cpp)
+add_executable(ArborX_DetailsTreeNodeLabeling.exe
+  tstDetailsTreeNodeLabeling.cpp
+  tstDetailsMutualReachabilityDistance.cpp
```
aprokop, line 166: Mutual reachability has nothing to do with labels, though. Should probably be its own test.

test/tstDetailsMutualReachabilityDistance.cpp
```diff
@@ -0,0 +1,167 @@
...
+#include "ArborX_EnableViewComparison.hpp"
+#include <ArborX_DetailsKokkosExtArithmeticTraits.hpp>
+#include <ArborX_DetailsMutualReachabilityDistance.hpp>
+#include <ArborX_LinearBVH.hpp>
+
+#include "BoostTest_CUDA_clang_workarounds.hpp"
+#include <boost/test/unit_test.hpp>
+
+template <typename DeviceType, typename T>
+auto toView(std::vector<T> const &v, std::string const &lbl = "")
```
aprokop, line 22: Given how we use this function in several tests, maybe we should move it some general test helpers file.

test/tstDetailsMutualReachabilityDistance.cpp
```diff
@@ -0,0 +1,167 @@
...
+template <class ExecutionSpace>
+auto compute_core_distances(ExecutionSpace exec_space,
+                            std::vector<ArborX::Point> const &points_host,
+                            int k)
+{
+  auto points = toView<ExecutionSpace>(points_host, "Test::points");
+
+  ARBORX_ASSERT(points.extent_int(0) >= k + 1);
+  using MemorySpace = typename ExecutionSpace::memory_space;
+  ArborX::BVH<MemorySpace> bvh{exec_space, points};
```
aprokop, line 43:
```suggestion
  ArborX::BVH<MemorySpace> bvh(exec_space, points);
```

test/tstDetailsMutualReachabilityDistance.cpp
```diff
@@ -0,0 +1,167 @@
...
+  Kokkos::View<float *, MemorySpace> distances(
+      Kokkos::view_alloc(Kokkos::WithoutInitializing, "Test::core_distances"),
+      bvh.size());
+  constexpr auto inf = KokkosExt::ArithmeticTraits::infinity<float>::value;
+  Kokkos::deep_copy(exec_space, distances, -inf);
```
aprokop, line 48: Can also be simply 0.f, as self-collisions are included.

aprokop (changes requested):
The definition of core distance includes the point itself. See Campello, Ricardo JGB, et al. "Hierarchical density estimates for data clustering, visualization, and outlier detection.", ACM TKDD 2015 (Section 3.1):
```
Definition 3.1 (Core Distance). The core distance of an object x_p ∈ X w.r.t. m_pts,
is the distance from x_p to its m_pts-nearest neighbor (including xp).
```
So the code and the tests need updates.

There is also some question about what is the exact definition of the mutual reachability distance. In Campello '15 it's just the max even when i = j. In McInnes '17, they set mutual reachability distance to 0 in that case. I have not carefully studied the side effects of this change, and it's outside of the scope of this PR. Just wanted to mention.

src/details/ArborX_DetailsMutualReachabilityDistance.hpp
```diff
@@ -0,0 +1,93 @@
...
+{
+
+template <class CoreDistances>
+struct MutualReachability
+{
+  CoreDistances _core_distances;
+  using value_type = typename CoreDistances::non_const_value_type;
+  using size_type = typename CoreDistances::memory_space::size_type;
+
+  value_type operator()(size_type i, size_type j, value_type distance_ij) const
```
aprokop, line 82: Should be `KOKKOS_FUNCTION`.

src/details/ArborX_DetailsMutualReachabilityDistance.hpp
dalg24: The definition of core distance does not mean nearest k neighbors include self collision, it means `k = min_core_size + 1` which is dealt with at call site.

test/CMakeLists.txt
dalg24: We should update the executable name indeed but I would still have then compiled into the same exe.  Do you have an issue with that?

test/tstDetailsMutualReachabilityDistance.cpp
dalg24: Will do

src/details/ArborX_DetailsMutualReachabilityDistance.hpp
dalg24: You are right

src/details/ArborX_DetailsMutualReachabilityDistance.hpp
aprokop: While I understand where you are coming from, I think this is still wrong. For a simple reason it is not guaranteed that the set of primitives and predicates match. If they don't, then the provided answer is different from what a user asked, as it will return more neighbors than expected. You could say that it's going to only be used in HDBSCAN, but I think given its convenience it's likely be used more widely, leading to errors.

dalg24: I don't follow you argument.
I am saying that if I do
```C++
BVH bvh{exec_space, primitives};
bvh.query(exec_space, NearestK{primitives, k}, MaxDistance{primitives, distances});
```
it feels more natural to me that `k` excludes self-collisions

dalg24: Note that I would use half traversal from #485 here

dalg24: We still don't quite agree on whether self collision should be included but we identified the source of confusion.
Andrey meant using `NearestK{geometries}` in general where `geometries` would not necessarily be the same as the `primitives` argument that was used to build `bvh`, whereas my intention was only a facility to solve nearest problem where they are the same and definitely not to provide a general facility.  We haven't reached a resolution yet but at least I think we understand each other better.  We may decide to accept temporarily one convention or the other to make progress on this and resume the argument later.

test/CMakeLists.txt
dalg24: I renamed the executable.

aprokop (commented):
Overall, this PR is in a good shape.

src/details/ArborX_DetailsMutualReachabilityDistance.hpp
```diff
@@ -0,0 +1,94 @@
...
+  using memory_space = typename Access::memory_space;
+  using size_type = typename memory_space::size_type;
+  template <class Predicate>
+  KOKKOS_FUNCTION void operator()(Predicate const &predicate, size_type i) const
+  {
+    size_type const j = getData(predicate);
+    using KokkosExt::max;
+    auto const distance_ij =
+        distance(Access::get(_primitives, i), Access::get(_primitives, j));
+    _distances(j) = max(_distances(j), distance_ij); // careful here
```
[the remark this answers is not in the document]
dalg24: Using knowledge that threads do not collaborate for the traversal associated with that predicate.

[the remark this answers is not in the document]
dalg24: Would need to be atomic update otherwise.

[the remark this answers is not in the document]
aprokop: I'd rather change that comment to state that, then.

aprokop (approved):

aprokop:
Related to #557.

aprokop:
SYCL query test timed out, not related.

PR #615, opened 2022-01-06
Restrict Thrust Clang workaround to Clang version less than 12

dalg24:
I certainly don't want to try all combinations of Clang and CUDA version but this change is required to build with Clang version greater equal 12 and CUDA 11.4


dalg24:
Nevermind it does not work, I get
```
/usr/local/cuda-11.4/include/cub/device/dispatch/../../agent/../block/radix_rank_sort_operations.cuh:124:20: error: explicit qualification required to use member 'ProcessFloatMinusZero' from dependent base class
        return BFE(ProcessFloatMinusZero(key), bit_start, num_bits);

PR #617, opened 2022-01-13
Implementation of the water tight ray-triangle intersection

wjge:
The algorithm is described in Woop, S. et al. (2013), Watertight Ray/Triangle Intersection, Journal of Computer Graphics Techniques Vol. 2(1), https://jcgt.org/published/0002/01/05/

### DIFF
--- a/src/details/ArborX_Ray.hpp
+++ b/src/details/ArborX_Ray.hpp
@@ -315,24 +315,33 @@ bool intersection(Ray const &ray, Triangle const &triangle, float &tmin,
   Vector const oB = makeVector(ray.origin(), triangle.b);
   Vector const oC = makeVector(ray.origin(), triangle.c);
 
+  // oA, oB, oB need to be normalized, otherwise they
+  // will scale with the problem size.
+  float const mag_oA = std::sqrt(dotProduct(oA, oA));
+  float const mag_oB = std::sqrt(dotProduct(oB, oB));
+  float const mag_oC = std::sqrt(dotProduct(oC, oC));
+
+  auto mag_bar = 3.0 / (mag_oA + mag_oB + mag_oC);
+
   Point A;
   Point B;
   Point C;
 
   // perform shear and scale of vertices
-  A[0] = oA[kx] - s[0] * oA[kz];
-  A[1] = oA[ky] - s[1] * oA[kz];
-  B[0] = oB[kx] - s[0] * oB[kz];
-  B[1] = oB[ky] - s[1] * oB[kz];
-  C[0] = oC[kx] - s[0] * oC[kz];
-  C[1] = oC[ky] - s[1] * oC[kz];
+  // normalized by mag_bar
+  A[0] = (oA[kx] - s[0] * oA[kz]) * mag_bar;
+  A[1] = (oA[ky] - s[1] * oA[kz]) * mag_bar;
+  B[0] = (oB[kx] - s[0] * oB[kz]) * mag_bar;
+  B[1] = (oB[ky] - s[1] * oB[kz]) * mag_bar;
+  C[0] = (oC[kx] - s[0] * oC[kz]) * mag_bar;
+  C[1] = (oC[ky] - s[1] * oC[kz]) * mag_bar;
 
   // calculate scaled barycentric coordinates
   float u = C[0] * B[1] - C[1] * B[0];
   float v = A[0] * C[1] - A[1] * C[0];
   float w = B[0] * A[1] - B[1] * A[0];
 
-  // fallback to edge test using double precision
+  // fallback to double precision
   if (u == 0 || v == 0 || w == 0)
   {
     u = (double)C[0] * B[1] - (double)C[1] * B[0];
@@ -352,7 +361,7 @@ bool intersection(Ray const &ray, Triangle const &triangle, float &tmin,
   // only one of the conditions is needed,
   // either (u < 0 || v < 0 || w < 0) or
   // (u > 0 || v > 0 || w > 0), for Back-facing culling.
-  auto const epsilon = 0.00001f;
+  float const epsilon = 0.0000001f;
   if ((u < -epsilon || v < -epsilon || w < -epsilon) &&
       (u > epsilon || v > epsilon || w > epsilon))
     return false;
@@ -369,8 +378,9 @@ bool intersection(Ray const &ray, Triangle const &triangle, float &tmin,
     float t = (u * A[2] + v * B[2] + w * C[2]) / det;
     tmax = t;
     tmin = t;
-    return tmax >= tmin;
+    return true;
   }
+
   // The ray is co-planar to the triangle.
   // Check the intersection with each edge
   // the rotate2D function is to make sure the ray-edge
@@ -404,16 +414,28 @@ bool intersection(Ray const &ray, Triangle const &triangle, float &tmin,
 
   if (ab_intersect || bc_intersect || ca_intersect)
   {
+    // When (1) the origin of the ray is within the triangle
+    // and (2) they ray is coplanar with the triangle, the
+    // intersection length is zero.
     if (tmin * tmax <= 0)
     {
       tmin = 0;
       tmax = 0;
     }
+    else
+    {
+      // need to separate tmin tmax >0 and <0 cases
+      // e.g., tmin = -2 and tmax = -1, but
+      // we want tmin = -1 and tmax = -2, when the
+      // ray travels backward
+      if (tmin < 0)
+        KokkosExt::swap(tmin, tmax);
+    }
     return true;
   }
 
   return false;
-}
+} // namespace Experimental
 
 KOKKOS_INLINE_FUNCTION
 bool intersects(Ray const &ray, Triangle const &triangle)
--- a/src/details/ArborX_Ray.hpp
+++ b/src/details/ArborX_Ray.hpp
@@ -17,6 +17,7 @@
 #include <ArborX_DetailsKokkosExtSwap.hpp>
 #include <ArborX_Point.hpp>
 #include <ArborX_Sphere.hpp>
+#include <ArborX_Triangle.hpp>
 
 #include <Kokkos_Macros.hpp>
 
@@ -196,6 +197,227 @@ bool intersects(Ray const &ray, Box const &box)
   float tmax;
   // intersects only if box is in front of the ray
   return intersection(ray, box, tmin, tmax) && (tmax >= 0.f);
+}
+
+// The function returns the index of the largest
+// component of the direction vector.
+KOKKOS_INLINE_FUNCTION int findLargestComp(Vector const &dir)
+{
+  int kz = 0;
+  for (int i = 1; i < 3; i++)
+  {
+    float compmax = std::fabs(dir[i - 1]);
+    if (std::fabs(dir[i]) > compmax)
+    {
+      compmax = dir[i];
+      kz = i;
+    }
+  }
+  return kz;
+}
+
+// Both the ray and the triangle were transformed beforehand
+// so that the ray is a unit vector along the z-axis (0,0,1),
+// and the triangle is transformed with the same matrix
+// (which is M in the paper). This function is called only
+// when the ray is co-planar to the triangle (with the
+// determinant being zero). The rotation by this function is
+// to prepare for the ray-edge intersection calculations in 2D.
+// The rotation is around the z-axis. For any point after
+// the rotation, its new x* equals its original length
+// with the correct sign, and the new y* = z. The current
+// implementation avoids explicitly defining rotation angles
+// and directions. The following ray-edge intersection will
+// be in the x*-y* plane.
+KOKKOS_INLINE_FUNCTION Point rotate2D(Point const &point)
+{
+  Point point_star;
+  float r = std::sqrt(point[0] * point[0] + point[1] * point[1]);
+  if (point[0] != 0)
+  {
+    point_star[0] = (point[0] > 0 ? 1 : -1) * r;
+  }
+  else
+  {
+    point_star[0] = (point[1] > 0 ? 1 : -1) * r;
+  }
+  point_star[1] = point[2];
+  point_star[2] = 0.0;
+  return point_star;
+}
+
+// The function is for ray-edge intersection
+// with the rotated ray along the z-axis and
+// the transformed and rotated triangle edges
+// The algorithm is described in
+// https://en.wikipedia.org/wiki/Line%E2%80%93line_intersection#Given_two_points_on_each_line_segment
+KOKKOS_INLINE_FUNCTION bool rayEdgeIntersect(Point const &edge_vertex_1,
+                                             Point const &edge_vertex_2,
+                                             float &t)
+{
+  float x3 = edge_vertex_1[0];
+  float y3 = edge_vertex_1[1];
+  float x4 = edge_vertex_2[0];
+  float y4 = edge_vertex_2[1];
+
+  float y1 = KokkosExt::min(y3, y4);
+  float y2;
+  if (y1 >= 0.f)
+  {
+    y2 = KokkosExt::max(y3, y4);
+  }
+  else
+  {
+    y2 = KokkosExt::min(y3, y4);
+  }
+
+  float det = y2 * (x3 - x4);
+
+  auto const epsilon = 0.00001f;
+  //  the ray is parallel to the edge if det == 0.0
+  //  When the ray overlaps the edge (x3==x4==0.0), it also returns false,
+  //  and the intersection will be captured by the other two edges.
+  if (det == 0)
+  {
+    return false;
+  }
+  t = (-x3 * (y3 - y4) + y3 * (x3 - x4)) / det * y2;
+
+  if (t >= 0)
+  {
+    float u = x3 * y2 / det;
+    if (u >= 0 - epsilon && u <= 1 + epsilon)
+    {
+      return true;
+    }
+  }
+  return false;
+}
+
+// The algorithm is described in
+// Watertight Ray/Triangle Intersection
+// [1] Woop, S. et al. (2013),
+// Journal of Computer Graphics Techniques Vol. 2(1)
+// The major difference is that here we return the intersection points
+// when the ray and the triangle is coplanar.
+// In the paper, they just need the boolean return.
+KOKKOS_INLINE_FUNCTION
+bool intersection(Ray const &ray, Triangle const &triangle, float &tmin,
+                  float &tmax)
+{
+  auto dir = ray.direction();
+  // normalize the direction vector by its largest component.
+  auto kz = findLargestComp(dir);
+  int kx = (kz + 1) % 3;
+  int ky = (kz + 2) % 3;
+
+  if (dir[kz] < 0)
+    KokkosExt::swap(kx, ky);
+
+  Vector s;
+
+  s[2] = 1.0f / dir[kz];
+  s[0] = dir[kx] * s[2];
+  s[1] = dir[ky] * s[2];
+
+  // calculate vertices relative to ray origin
+  Vector const oA = makeVector(ray.origin(), triangle.a);
+  Vector const oB = makeVector(ray.origin(), triangle.b);
+  Vector const oC = makeVector(ray.origin(), triangle.c);
+
+  Point A;
+  Point B;
+  Point C;
+
+  // perform shear and scale of vertices
+  A[0] = oA[kx] - s[0] * oA[kz];
+  A[1] = oA[ky] - s[1] * oA[kz];
+  B[0] = oB[kx] - s[0] * oB[kz];
+  B[1] = oB[ky] - s[1] * oB[kz];
+  C[0] = oC[kx] - s[0] * oC[kz];
+  C[1] = oC[ky] - s[1] * oC[kz];
+
+  // calculate scaled barycentric coordinates
+  float u = C[0] * B[1] - C[1] * B[0];
+  float v = A[0] * C[1] - A[1] * C[0];
+  float w = B[0] * A[1] - B[1] * A[0];
+
+  // fallback to edge test using double precision
+  if (u == 0 || v == 0 || w == 0)
+  {
+    u = (double)C[0] * B[1] - (double)C[1] * B[0];
+    v = (double)A[0] * C[1] - (double)A[1] * C[0];
+    w = (double)B[0] * A[1] - (double)B[1] * A[0];
+  }
+
+  constexpr auto inf = KokkosExt::ArithmeticTraits::infinity<float>::value;
+  tmin = inf;
+  tmax = -inf;
+
+  // The following 'if' statement work
+  // regardless of the facing of the triangle.
+  // In another word, 'Back-face culling' is not supported.
+  // Back-facing culling is to check whether
+  // a surface is 'visible' to a ray, which requires
+  // consistent definition of the facing of triangles.
+  if ((u < 0 || v < 0 || w < 0) && (u > 0 || v > 0 || w > 0))
+    return false;
+
+  // calculate determinant
+  float det = u + v + w;
+
+  A[2] = s[2] * oA[kz];
+  B[2] = s[2] * oB[kz];
+  C[2] = s[2] * oC[kz];
+
+  if (det != 0)
+  {
+    float t = (u * A[2] + v * B[2] + w * C[2]) / det;
+    tmax = t;
+    tmin = t;
+    return tmax >= tmin;
+  }
+  // The ray is co-planar to the triangle.
+  // Check the intersection with each edge
+  // the rotate2D function is to make sure the ray-edge
+  // intersection check is at the plane where ray and edges
+  // are at.
+  auto A_star = rotate2D(A);
+  auto B_star = rotate2D(B);
+  auto C_star = rotate2D(C);
+
+  float t_ab = inf;
+  bool ab_intersect = rayEdgeIntersect(A_star, B_star, t_ab);
+  if (ab_intersect)
+  {
+    tmin = t_ab;
+    tmax = t_ab;
+  }
+  float t_bc = inf;
+  bool bc_intersect = rayEdgeIntersect(B_star, C_star, t_bc);
+  if (bc_intersect)
+  {
+    tmin = KokkosExt::min(tmin, t_bc);
+    tmax = KokkosExt::max(tmax, t_bc);
+  }
+  float t_ca = inf;
+  bool ca_intersect = rayEdgeIntersect(C_star, A_star, t_ca);
+  if (ca_intersect)
+  {
+    tmin = KokkosExt::min(tmin, t_ca);
+    tmax = KokkosExt::max(tmax, t_ca);
+  }
+
+  return (ab_intersect || bc_intersect || ca_intersect);
+}
+
+KOKKOS_INLINE_FUNCTION
+bool intersects(Ray const &ray, Triangle const &triangle)
+{
+  float tmin;
+  float tmax;
+  // intersects only if triangle is in front of the ray
+  return intersection(ray, triangle, tmin, tmax) && (tmax >= 0.f);
 }
 
 // Solves a*x^2 + b*x + c = 0.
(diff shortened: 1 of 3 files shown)


aprokop:
Alternative to #604.

src/details/ArborX_Ray.hpp
```diff
@@ -198,6 +199,216 @@ bool intersects(Ray const &ray, Box const &box)
...
+  C[0] = oC[kx] - s[0] * oC[kz];
+  C[1] = oC[ky] - s[1] * oC[kz];
+
+  // calculate scaled barycentric coordinates
+  float u = C[0] * B[1] - C[1] * B[0];
+  float v = A[0] * C[1] - A[1] * C[0];
+  float w = B[0] * A[1] - B[1] * A[0];
+
+  // fallback to edge test using double precision
+  if (u == 0.0f || v == 0.0f || w == 0.0f)
```
aprokop, line 332: According to the paper:
```
an algorithm that works under all circumstances can easily be obtained
through a fallback of the 2D edge test to double precision if and only if one of
the single-precision tests evaluates to 0.
```
So only one of them can be 0. And use plain `0` instead of `0.0f` for comparison.

src/details/ArborX_Ray.hpp
```diff
@@ -198,6 +199,216 @@ bool intersects(Ray const &ray, Box const &box)
...
+  {
+    double CxBy = (double)C[0] * (double)B[1];
+    double CyBx = (double)C[1] * (double)B[0];
+    u = (float)(CxBy - CyBx);
```
aprokop, line 336:
```suggestion
    u = (double)C[0] * B[1] - (double)C[1] * B[0];
```
Same for the other two.

src/details/ArborX_Ray.hpp
```diff
@@ -198,6 +199,216 @@ bool intersects(Ray const &ray, Box const &box)
...
+    double ByAx = (double)B[1] * (double)A[0];
+    w = (float)(BxAy - ByAx);
+  }
+
+  constexpr auto inf = KokkosExt::ArithmeticTraits::infinity<float>::value;
+  tmin = inf;
+  tmax = -inf;
+
+  // depending on the facing of the triangle
+  if ((u < 0.0f || v < 0.0f || w < 0.0f) && (u > 0.0f || v > 0.0f || w > 0.0f))
```
aprokop, line 352:
```suggestion
  if ((u < 0 || v < 0 || w < 0) && (u > 0 || v > 0 || w > 0))
```

src/details/ArborX_Ray.hpp
```diff
@@ -198,6 +199,216 @@ bool intersects(Ray const &ray, Box const &box)
...
+    return false;
+
+  // calculate determinant
+  float det = u + v + w;
+
+  A[2] = s[2] * oA[kz];
+  B[2] = s[2] * oB[kz];
+  C[2] = s[2] * oC[kz];
+
+  if (det != 0.0f)
```
aprokop, line 362:
```suggestion
  if (det != 0)
```

src/details/ArborX_Ray.hpp
```diff
@@ -198,6 +199,216 @@ bool intersects(Ray const &ray, Box const &box)
...
+    float t = (u * A[2] + v * B[2] + w * C[2]) / det;
+    tmax = t;
+    tmin = t;
+    return tmax >= tmin;
+  }
+  // The ray is co-planar to the triangle.
+  // Check the intersection with each edgel
+  // the rotate2D function is to make sure the ray-edge
+  // intersection check is at the plane where ray and edges
+  // are at.
```
aprokop, line 373: Still not sure if this is right. There is nothing like this in the paper, and they claim that they still detect ray-triangle edge hit. Have you tried running without?

My thinking is that if `det` is 0, that means that `(u,v,w) = (0,0,0)` (because we already tested them for the same-signess), which are not valid barycentric coordinates. So, I think a ray does not intersect with the triangle at all. So, I think
```c++
  if (det == 0)
    return false;

  float t = (u * A[2] + v * B[2] + w * C[2]) / det
  tmax = t;
  tmin = t;
  return tmax >= tmin;
```
may work.

aprokop (commented):
Partial review

src/details/ArborX_Ray.hpp
aprokop: Hmm, does not actually work. And in the paper they actually state:
```
If this determinant is zero, the ray is co-planar to the triangle and we assume a miss.
```
I guess their definition of "hitting exactly on the edge" means something else; likely, only for non-planar intersections.

Rombur: That's the exact same problem in the other algorithm. The algorithm detects that you hit an edge except if the ray is co-planar.

src/details/ArborX_Ray.hpp
```diff
@@ -198,6 +199,216 @@ bool intersects(Ray const &ray, Box const &box)
...
+    }
+  }
+  int kx = kz + 1;
+  if (kx == 3)
+    kx = 0;
+  int ky = kx + 1;
+  if (ky == 3)
+    ky = 0;
+
+  if (dir[kz] < 0.0f)
```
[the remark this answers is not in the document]
wjge: A decision has to be made first on whether we want to normalize all the rays by its largest component at the initialization.

src/details/ArborX_Ray.hpp
```diff
@@ -198,6 +199,216 @@ bool intersects(Ray const &ray, Box const &box)
...
+    KokkosExt::swap(kx, ky);
+
+  Vector s;
+
+  s[2] = 1.0f / dir[kz];
+  s[0] = dir[kx] * s[2];
+  s[1] = dir[ky] * s[2];
```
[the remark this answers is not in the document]
wjge: division vs multiplication. Does it matter any more?

src/details/ArborX_Ray.hpp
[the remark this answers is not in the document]
aprokop: It makes sense from both the accuracy point of view, and the fact the that the z component may be zero.

src/details/ArborX_Ray.hpp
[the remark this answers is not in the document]
aprokop: This is not about * vs /. Rather, a single line of code vs five.

src/details/ArborX_Ray.hpp
```diff
@@ -198,6 +199,216 @@ bool intersects(Ray const &ray, Box const &box)
   return intersection(ray, box, tmin, tmax) && (tmax >= 0.f);
 }
 
+// rotate to the x-axis
+KOKKOS_INLINE_FUNCTION Point rotate2D(Point const &point)
```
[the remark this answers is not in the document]
wjge: The function rotates a 3-D plane which the ray and triangles are coplanar to a 2-D plane, so that the below algorithm can be used to find the intersection points.
https://en.wikipedia.org/wiki/Line%E2%80%93line_intersection#Given_two_points_on_each_line_segment
There might be other better/standard ways to do it.

src/details/ArborX_Ray.hpp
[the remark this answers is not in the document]
wjge: The issue is kx, ky, kz need to be stored. And they are not used for intersections with box or sphere.

src/details/ArborX_Ray.hpp
```diff
@@ -198,6 +199,216 @@ bool intersects(Ray const &ray, Box const &box)
...
+      compmax = dir[i];
+      kz = i;
+    }
+  }
+  int kx = kz + 1;
+  if (kx == 3)
+    kx = 0;
+  int ky = kx + 1;
+  if (ky == 3)
+    ky = 0;
```
aprokop, line 298:
```suggestion
  // normalizing the direction vector by its largest component.
  // if precalculated, kx, ky, kz need to be stored.
  int kz = 2;
  for (int i = 0; i < 2; ++i)
    if (std::fabs(dir[i]) > std::fabs(dir[kz]))
      kz = i;
 int kx = (kz + 1) % 3;
 int ky = (kz + 2) % 3;
```

src/details/ArborX_Ray.hpp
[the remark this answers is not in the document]
aprokop: I don't know if they need to be stored. The pre-calculations are just a couple operations, and should be cheap compared to everything else in the function.

aprokop:
Just a thought: maybe look at how Embree ([here](https://github.com/embree/embree/blob/master/kernels/geometry/triangle_intersector_pluecker.h) and other `triangle_intersector*.h` files) implements it.

src/details/ArborX_Ray.hpp
```diff
@@ -198,6 +199,216 @@ bool intersects(Ray const &ray, Box const &box)
...
+    costheta = std::fabs(point[0]) / r;
+    sintheta = std::fabs(point[1]) / r;
+  }
+  point_star[0] = point[0] * costheta + point[1] * sintheta;
+  point_star[1] = point[2];
+  point_star[2] = 0.0;
+  return point_star;
+}
+
+// modified from Bruno's PR #604
```
Rombur, line 220: Why did you change it?

aprokop:
As a side note, I want to point out two things:
- While the algorithm in the paper is guaranteed to be watertight, it is only for the case when `det = 0` returns false. The code in this PR that does edge intersection in this PR may not be watertight.
- There is also a question of intersection of a ray with a bounding box that may result in false negatives. See [this](https://jcgt.org/published/0002/02/02/) paper for example (I think they fix it by extending each box by 2 ulp in each direction). So may need some more work for robustness.

src/details/ArborX_Ray.hpp
```diff
@@ -198,6 +199,216 @@ bool intersects(Ray const &ray, Box const &box)
...
+  float costheta = 0.0f;
+  float sintheta = 0.0f;
+  if (r != 0.0f)
+  {
+    costheta = std::fabs(point[0]) / r;
+    sintheta = std::fabs(point[1]) / r;
+  }
+  point_star[0] = point[0] * costheta + point[1] * sintheta;
+  point_star[1] = point[2];
+  point_star[2] = 0.0;
```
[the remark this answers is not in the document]
wjge: I don't know how to express it in a more mathematical way. It is really simpler than it looks.

The triangle won't deform because any triangle reaching this step is coplanar with z-axis. 
If the vertices are projected onto the x-y plane they will be on the same line that also passes the origin.

For any point, after rotation, its new x equals its original length.
let's say one point A in x>0, y>0, the new x for A* is r, 
another point B in x<0, y>0, the new x for B* is -r, 
another point C in x>0, y<0, the new x for C* is r.

In this way, I don't need to specify the rotation direction. I just use the above calculation to replace if(x?0,y?0) statements. I agree this might not be the best way to do it.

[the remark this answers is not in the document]
aprokop: The paper's algorithm rotates the ray so that it becomes `(0, 0, 1)`. I think vertices `A`, `B` and `C` are rotated correspondingly. So I think it's correct to say that the rotated triangle will be co-planar to z-axis. And @wjge calls `rotate2D` on `A` and friends.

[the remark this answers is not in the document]
wjge: I am sorry I used the wrong examples in the reply, but I hope you get what I mean. I need to at least at more comments to explain it.

src/details/ArborX_Triangle.hpp
```diff
@@ -0,0 +1,21 @@
+#ifndef ARBORX_TRIANGLE_HPP
+#define ARBORX_TRIANGLE_HPP
+
+#include <ArborX_Box.hpp>
+#include <ArborX_DetailsAlgorithms.hpp>
```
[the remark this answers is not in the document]
wjge: Do we need to include the 'expand' function in the class?

KOKKOS_INLINE_FUNCTION void expand(Box &box, Triangle const &triangle)
{
  using Details::expand;
  expand(box, triangle.a);
  expand(box, triangle.b);
  expand(box, triangle.c);
}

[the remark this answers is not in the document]
dalg24: All you have in that header file is a definition for the `Triangle` struct so no.

src/details/ArborX_Ray.hpp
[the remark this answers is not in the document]
wjge: This change will fail one of the tests: 
intersects(Ray{{1,2,3},{0,-2,-3}}, unit_triangle)

When I print out s, the values are the same. s[0]=0.666667, s[1]=-0.000000, s[2]=-1.201850

src/details/ArborX_Ray.hpp
wjge: The suggestion failed one test. I put the original code for finding the index for the largest component into a function.

src/details/ArborX_Ray.hpp
```diff
@@ -198,6 +199,226 @@ bool intersects(Ray const &ray, Box const &box)
...
+    u = (double)C[0] * B[1] - (double)C[1] * B[0];
+    v = (double)A[0] * C[1] - (double)A[1] * C[0];
+    w = (double)B[0] * A[1] - (double)B[1] * A[0];
+  }
+
+  constexpr auto inf = KokkosExt::ArithmeticTraits::infinity<float>::value;
+  tmin = inf;
+  tmax = -inf;
+
+  // depending on the facing of the triangle
```
Rombur, line 361: This comment is not very clear. Depending on the facing of the triangle, then what?

wjge: I didn't comment it more because 'Back-face culling' is not supported at this moment.
Back-facing culling is to check whether a surface is 'visible' to a ray, which requires consistent definition of the facing of triangles.
Once the facing of triangle is defined, only one of the conditions is needed, either (u < 0 || v < 0 || w < 0) or (u > 0 || v > 0 || w > 0), for Back-facing culling. That's why the condition is "depending on the facing of the triangle". 
I edited this comment in the new commit.

src/details/ArborX_Ray.hpp
```diff
@@ -198,6 +199,226 @@ bool intersects(Ray const &ray, Box const &box)
...
+
+  auto const epsilon = 0.00001f;
+  //  the ray is parallel to the edge if det == 0.0
+  //  When the ray overlaps the edge (x3==x4==0.0), it also returns false,
+  //  and the intersection will be captured by the other two edges.
+  if (det == 0)
+  {
+    return false;
+  }
+  // t = ((x1 - x3) * (y3 - y4) - (y1 - y3) * (x3 - x4)) / det;
```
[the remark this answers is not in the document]
wjge: These are just from the original code of the other PR for more general cases.
I commented these out in the new commits.

Rombur (approved):
Looks good. Can you clean the history.

src/details/ArborX_Ray.hpp
```diff
@@ -198,6 +199,231 @@ bool intersects(Ray const &ray, Box const &box)
...
+  if (point[0] != 0)
+  {
+    point_star[0] = (point[0] > 0 ? 1 : -1) * r;
+  }
+  else
+  {
+    point_star[0] = (point[1] > 0 ? 1 : -1) * r;
+  }
+  point_star[1] = point[2];
+  point_star[2] = 0.0;
```
aprokop, line 245:
```suggestion
  point_star[2] = 0.f;
```

src/details/ArborX_Ray.hpp
```diff
@@ -198,6 +199,231 @@ bool intersects(Ray const &ray, Box const &box)
...
+
+  float y2;
+  if (KokkosExt::min(y3, y4) >= 0.f)
+  {
+    y2 = KokkosExt::max(y3, y4);
+  }
+  else
+  {
+    y2 = KokkosExt::min(y3, y4);
+  }
```
aprokop, line 271: This looks a bit weird. Is the assumption here that `y3` and `y4` have the same sign?

src/details/ArborX_Ray.hpp
```diff
@@ -198,6 +199,231 @@ bool intersects(Ray const &ray, Box const &box)
...
+
+  float det = y2 * (x3 - x4);
+
+  auto const epsilon = 0.00001f;
```
aprokop, line 275: Magic constant. Would be nice to explain why we need it, and what should be a proper value.

src/details/ArborX_Ray.hpp
```diff
@@ -198,6 +199,231 @@ bool intersects(Ray const &ray, Box const &box)
...
+  //  the ray is parallel to the edge if det == 0.0
+  //  When the ray overlaps the edge (x3==x4==0.0), it also returns false,
+  //  and the intersection will be captured by the other two edges.
+  if (det == 0)
+  {
+    return false;
+  }
+  t = (-x3 * (y3 - y4) + y3 * (x3 - x4)) / det * y2;
```
aprokop, line 283: Isn't this the same as
```suggestion
  t = (x3 * y4 - x4 * y3) / det * y2;
```
?

src/details/ArborX_Triangle.hpp
```diff
@@ -0,0 +1,21 @@
...
+
+#include <ArborX_Point.hpp>
+
+namespace ArborX
+{
+namespace Experimental
+{
+// need to add a protection that
+// the points are not on the same line.
+struct Triangle
```
aprokop, line 12: Note to self: 36 byte alignment. Not great, but don't know if playing with would make any difference.

aprokop (approved):
Looks fine, thanks @wjge and @Rombur. I only have few minor comments.

src/details/ArborX_Ray.hpp
```diff
@@ -198,6 +199,231 @@ bool intersects(Ray const &ray, Box const &box)
...
+  {
+    float compmax = std::fabs(dir[i - 1]);
+    if (std::fabs(dir[i]) > compmax)
+    {
+      compmax = dir[i];
+      kz = i;
+    }
+  }
+  return kz;
+}
```
dalg24, line 217: Consider `dir = {3,1,2}`, your function will return `2` which I expect is not what you wanted is it?

dalg24 (dismissed):

src/details/ArborX_Ray.hpp
aprokop: Yeah, and also problems with signs. It should be 
```suggestion
KOKKOS_INLINE_FUNCTION int findLargestComp(Vector const &dir)
{
  int kz = 0;
  auto max = std::abs(dir[0]);
  for (int i = 1; i < 3; i++)
  {
    auto f = std::fabs(dir[i]);
    if (std::fabs(f) > max)
    {
      max = f;
      kz = i;
    }
  }
  return kz;
}
```

src/details/ArborX_Ray.hpp
wjge: It depends on the size of the problem as briefly discussed in the previous PR.
See the new tests and the new epsilon added in the commit.

wjge: If I further increase the size range, the current epsilon (for u v w and det) will fail.

wjge: Bruno found in the previous PR that it was due to precisions.
I think the roll-back precision in this algorithm doesn't work as it claimed to be, at least the roll-back itself needs an epsilon to check if it is needed.
But if you check the numbers when scaling the problem in size, the algorithm is not scaling with it with the precisions/if checks.

src/details/ArborX_Ray.hpp
```diff
@@ -204,15 +204,20 @@ bool intersects(Ray const &ray, Box const &box)
...
 {
   int kz = 0;
+
+  auto max = std::abs(dir[0]);
+
   for (int i = 1; i < 3; i++)
   {
-    float compmax = std::fabs(dir[i - 1]);
-    if (std::fabs(dir[i]) > compmax)
+    auto f = std::fabs(dir[i]);
```
masterleinad, line 212: We can't use namespace `std` in code that runs on the device portably.

dalg24: To be clear we do not request this to be changed

src/details/ArborX_Ray.hpp
```diff
@@ -198,6 +199,231 @@ bool intersects(Ray const &ray, Box const &box)
...
+
+  // 'Back-face culling' is not supported.
+  // Back-facing culling is to check whether
+  // a surface is 'visible' to a ray, which requires
+  // consistent definition of the facing of triangles.
+  // Once the facing of triangle is defined,
+  // only one of the conditions is needed,
+  // either (u < 0 || v < 0 || w < 0) or
+  // (u > 0 || v > 0 || w > 0), for Back-facing culling.
+  auto const epsilon = 0.00001f;
```
dalg24, line 355: Why do you have an epsilon?  The original code does not.

wjge: the value of u, v, w will scale with the size of the triangle.
I am still figuring out how to normalize them so that the u, v, w won't change for the test I added in the latest commit (line 533 in tstRay.cpp).

src/details/ArborX_Ray.hpp
```diff
@@ -198,6 +199,231 @@ bool intersects(Ray const &ray, Box const &box)
...
+  float v = A[0] * C[1] - A[1] * C[0];
+  float w = B[0] * A[1] - B[1] * A[0];
+
+  // fallback to edge test using double precision
+  if (u == 0 || v == 0 || w == 0)
+  {
+    u = (double)C[0] * B[1] - (double)C[1] * B[0];
+    v = (double)A[0] * C[1] - (double)A[1] * C[0];
+    w = (double)B[0] * A[1] - (double)B[1] * A[0];
+  }
```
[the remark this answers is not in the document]
wjge: I am still thinking about this part, u==0 v==0 w==0 is in float precision.
Maybe we want to fallback to double precision by default.

src/details/ArborX_Triangle.hpp
```diff
@@ -0,0 +1,21 @@
...
+#define ARBORX_TRIANGLE_HPP
+
+#include <ArborX_Point.hpp>
+
+namespace ArborX
+{
+namespace Experimental
+{
+// need to add a protection that
+// the points are not on the same line.
```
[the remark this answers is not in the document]
wjge: I mean the triangle should not be a single line when constructed.
The ray-triangle algorithm doesn't cover such cases.

src/details/ArborX_Ray.hpp
```diff
@@ -198,6 +199,231 @@ bool intersects(Ray const &ray, Box const &box)
...
+    tmax = KokkosExt::max(tmax, t_ca);
+  }
+
+  if (ab_intersect || bc_intersect || ca_intersect)
+  {
+    if (tmin * tmax <= 0)
+    {
+      tmin = 0;
+      tmax = 0;
+    }
```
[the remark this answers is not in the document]
dalg24: Add a comment plz

src/details/ArborX_Ray.hpp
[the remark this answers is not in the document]
dalg24: I am just saying the comment is bad (and yes I know that is the comment from the original code)

wjge:
I am still not happy about the algorithm because u, v, w are not normalized by the area of the triangle. This is the reason why we need a magic epsilon. See the reference here:

https://www.scratchapixel.com/lessons/3d-basic-rendering/ray-tracing-rendering-a-triangle/barycentric-coordinates

src/details/ArborX_Ray.hpp
masterleinad: Can't we normalize coordinates by the maximum of the difference between vertices and ray origin and rescale the ray length to be one?

src/details/ArborX_Ray.hpp
```diff
@@ -198,6 +199,231 @@ bool intersects(Ray const &ray, Box const &box)
...
+// when the ray is co-planar to the triangle (with the
+// determinant being zero). The rotation by this function is
+// to prepare for the ray-edge intersection calculations in 2D.
+// The rotation is around the z-axis. For any point after
+// the rotation, its new x* equals its original length
+// with the correct sign, and the new y* = z. The current
+// implementation avoids explicitly defining rotation angles
+// and directions. The following ray-edge intersection will
+// be in the x*-y* plane.
+KOKKOS_INLINE_FUNCTION Point rotate2D(Point const &point)
```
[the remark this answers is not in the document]
wjge: The name is too long. And it is a rotation in a simplified way, not a projection.

We shall move all sub-functions in ray tracing to another namespace. Not sure if we want to do it now.

src/details/ArborX_Ray.hpp
```diff
@@ -198,6 +199,231 @@ bool intersects(Ray const &ray, Box const &box)
...
+  A[2] = s[2] * oA[kz];
+  B[2] = s[2] * oB[kz];
+  C[2] = s[2] * oC[kz];
+
+  if (det < -epsilon || det > epsilon)
+  {
+    float t = (u * A[2] + v * B[2] + w * C[2]) / det;
+    tmax = t;
+    tmin = t;
+    return tmax >= tmin;
```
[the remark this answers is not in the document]
wjge: What do you mean?
This definition follows the box-ray intersection.

wjge:
I get something working, combining with @masterleinad 's suggestion.
Basic idea is to use average of the distances between vertices and ray origin to normalize.
Need to clean up the code.

src/details/ArborX_Ray.hpp
```diff
@@ -198,6 +199,244 @@ bool intersects(Ray const &ray, Box const &box)
...
+  Vector const oB = makeVector(ray.origin(), triangle.b);
+  Vector const oC = makeVector(ray.origin(), triangle.c);
+
+  // oA, oB, oB need to be normalized, otherwise they
+  // will scale with the problem size.
+  float const mag_oA = std::sqrt(dotProduct(oA, oA));
+  float const mag_oB = std::sqrt(dotProduct(oB, oB));
+  float const mag_oC = std::sqrt(dotProduct(oC, oC));
+
+  auto mag_bar = 3.0 / (mag_oA + mag_oB + mag_oC);
```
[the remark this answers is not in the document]
wjge: it is actually the inverse of the average of the magnitudes/lengths.
'bar' stands for average.

src/details/ArborX_Ray.hpp
```diff
@@ -198,6 +199,244 @@ bool intersects(Ray const &ray, Box const &box)
...
+  bool ca_intersect = rayEdgeIntersect(C_star, A_star, t_ca);
+  if (ca_intersect)
+  {
+    tmin = KokkosExt::min(tmin, t_ca);
+    tmax = KokkosExt::max(tmax, t_ca);
+  }
+
+  if (ab_intersect || bc_intersect || ca_intersect)
+  {
+    // When (1) the origin of the ray is within the triangle
```
Rombur, line 417: Shouldn't it be:
```suggestion
    // When (1) the origin of the ray is on one of the edges of the triangle
```

wjge: We need to define the cases which the ray origin is inside of the triangle, as well.

test/tstRay.cpp
```diff
@@ -375,6 +376,175 @@ BOOST_AUTO_TEST_CASE(ray_sphere_intersection,
...
+  using ArborX::Experimental::Triangle;
+
+  constexpr Triangle unit_triangle{{0, 0, 0}, {1, 0, 0}, {0, 1, 0}};
+  constexpr Triangle narrow_triangle{{0.5, 0.5, 0}, {0.24, 0.74, 0}, {0, 1, 0}};
+
+  auto const sqrtf_3 = std::sqrt(3.f);
+  auto const sqrtf_2 = std::sqrt(2.f);
+
+  // clang-format off
+  ARBORX_TEST_RAY_TRIANGLE_INTERSECTION((Ray{{0.01, 0.2, -1.0}, {0, 0, 1}}), unit_triangle, 1.f, 1.f);
```
Rombur, line 473: Could you add comments where the ray is supposed to hit like it's done above.

Rombur:
Retest this please

Rombur:
Retest this please

dalg24 (approved):

masterleinad (approved):

PR #618, opened 2022-01-20
Add NVHPC nightly build

dalg24:


docker/Dockerfile.nvhpc
```diff
@@ -0,0 +1,53 @@
+ARG BASE=nvcr.io/nvidia/nvhpc:21.9-devel-cuda11.4-ubuntu20.04
+FROM $BASE
+
+ARG NPROCS=4
+
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -yq \
+        libboost-program-options-dev \
+        libboost-test-dev \
+        libbenchmark-dev
```
masterleinad, line 10: What boost version does this correspond to? Are we running into the same problems as in https://github.com/arborx/ArborX/issues/168?
We even print
```
Boost versions 1.69 to 1.74 are known to yield build issues with NVCC
```
So we should probably just try a newer `Boost` version.

PR #625, opened 2022-02-07
Revert "Remove __SYCL_DEVICE_ONLY__ guards for ptintf"

masterleinad:
This reverts commit d54e117fc0b7b6a6a36a4e1df0f66f58a244dc49 (part of #614). Corresponding to https://github.com/kokkos/kokkos/pull/4732.


aprokop (approved):
We revert because the fact that `printf` worked was accidental, and it became broken again.

PR #654, opened 2022-03-10
Add Ray-Box nearest/intersects test

masterleinad:
The only new thing is to implement
````
float distance(Ray const &ray, Box const &box)
````

I'll add some more test cases.

### CODE REVIEW

**Comment by wjge** [comment]
Could you describe what you are trying to implement?

**Comment by masterleinad** [comment]
> Could you describe what you are trying to implement?

The idea is to have a raytracer proxy at some point and for that, I typically only need a first intersection, i.e. a nearest search.

File: src/details/ArborX_Ray.hpp [hunk 1]
```diff
@@ -198,6 +198,16 @@ bool intersects(Ray const &ray, Box const &box)
...
 }
 
+KOKKOS_INLINE_FUNCTION
+float distance(Ray const &ray, Box const &box)
+{
+  float tmin;
+  float tmax;
+  bool intersects = intersection(ray, box, tmin, tmax) && (tmax >= 0.f);
+  return intersects ? (tmin > 0.f ? tmin : 0.f)
+                    : KokkosExt::ArithmeticTraits::infinity<float>::value;
```
aprokop, line 208: This is not what a distance between ray and box is. IMO, the distance between Ray and Box is always well defined and finite. Instead, this looks like "Distance from the ray origin to the box if they intersect, and infinity otherwise". So I would call it something else.

dalg24: I don't see the problem with the definition proposed here.  `inf` denotes "no distance".  What am I missing?

masterleinad: I think @aprokop doesn't like that the definition here doesn't match the default "closest distance between two objects" which is not what I need here.

aprokop: Yes. `distance` for me is the distance between two geometric objects. As such, distance from a valid ray to a valid box is always finite.

masterleinad: Any suggestions on how I can specify what I need/want using our current interface then?

aprokop: It's hard for me to say what you and @dalg24  are doing without seeing the code. But you could still use the function, just not name is `distance`.

masterleinad: For ray-tracing, I need to know what the first object is a ray intersects and a `nearest` query is doing exactly that. More precisely, I don't want the distance between a Ray and an object, but the distance between a point and an object with respect to a given direction.
The `nearest` traversal uses `distance` to order matches and thus I need to define `distance` in order to use it.

aprokop: So I understand how that would allow you to find the closest object that ray intersects with. But I don't see how that would allow you to order all objects that a ray intersects with in increasing order distance. I thought for that you would need additional structures, as you would essentially need to maintain possibly unbounded number of intersections.

I also think we need to start using the proper terminology to describe different algorithms. I think following [this](https://onlinelibrary.wiley.com/doi/abs/10.1111/cgf.142662) paper could be a good choice. In Section 6.1, they establish several terms:
- First hit traversal (closest hit test)
  Finding the nearest object in the direction of a ray from its origin
- Any-hit traversal (occlusion test)
  Finding any object in a specific direction from a ray origin (and within a specified range)
- Multi-hit traversal
  Finding one or more primitives that intersect with a ray (k closest intersections from a viewpoint or all intersections)

So, in these terms, is the problem you are considering here first hit traversal, or multi-hit traversal?

masterleinad: At this point, I want "First hit traversal". I don't think we need anything else for a reasonable proxy.

aprokop: OK, that makes more sense. I thought you were considering the problem that the university team brought up a few months ago, where you wanted to do all hits but ordered.

test/tstQueryTreeCallbacksRay.cpp
```diff
@@ -0,0 +1,156 @@
...
+ * the LICENSE file in the top-level directory.                             *
+ *                                                                          *
+ * SPDX-License-Identifier: BSD-3-Clause                                    *
+ ****************************************************************************/
+
+#include <ArborX.hpp>
+
+#include <boost/test/unit_test.hpp>
+
+#include <numeric>
```
dalg24, line 16: What do you use from this header?

test/tstQueryTreeCallbacksRay.cpp
```diff
@@ -0,0 +1,156 @@
...
+  using memory_space = typename DeviceType::memory_space;
+  static KOKKOS_FUNCTION int size(Boxes<DeviceType> const &boxes)
+  {
+    return boxes.elements.size();
+  }
+  static KOKKOS_FUNCTION auto get(Boxes<DeviceType> const &boxes, int i)
+  {
+    return boxes.elements(i);
+  }
+};
```
dalg24, line 57: Why did you define `Boxes` with your own access traits?  We support `Kokkos::View<T*>` natively.

src/ArborX.hpp
```diff
@@ -23,6 +23,7 @@
 #include <ArborX_LinearBVH.hpp>
 #include <ArborX_Point.hpp>
 #include <ArborX_Predicates.hpp>
+#include <ArborX_Ray.hpp>
```
dalg24, line 26: Please don't.  This was intentional.  You had to explicitly include it to get support for raytracing.
I am not necessarily opposed to revisit but it does not belong here.

test/CMakeLists.txt
```diff
@@ -121,6 +121,7 @@ foreach(_test Callbacks Degenerate ManufacturedSolution ComparisonWithBoost)
 endforeach()
 add_executable(ArborX_QueryTree.exe
   ${ARBORX_TEST_QUERY_TREE_SOURCES}
+  tstQueryTreeCallbacksRay.cpp
```
dalg24, line 124: File name is not good.

Also wondering if we should add a dedicated executable for it.
I suppose we could settle that later.  @Rombur @wjge what do you think?

test/tstQueryTreeCallbacksRay.cpp
```diff
@@ -0,0 +1,156 @@
...
+  }
+  static KOKKOS_FUNCTION auto
+  get(BoxesIntersectedByRay<DeviceType> const &nearest_boxes, int i)
+  {
+    return intersects(nearest_boxes.rays(i));
+  }
+};
+} // namespace ArborX
+
+BOOST_AUTO_TEST_SUITE(RayCallbacks)
```
dalg24, line 92: Needs update to reflect better name for the file.

dalg24 (changes requested):

test/tstRay.cpp
```diff
@@ -13,6 +13,176 @@
 
 #include <boost/test/unit_test.hpp>
 
+BOOST_AUTO_TEST_CASE(distance_box)
```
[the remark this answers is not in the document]
masterleinad: Yes, I just copied what we have. If you don't want to have both I would rather keep the distance tests and remove the intersection tests since they test more.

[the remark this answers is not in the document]
dalg24: You mean keep the original one?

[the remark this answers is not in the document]
masterleinad: No, the new checking distances. `inf` means "doesn't intersect", all other numbers mean "intersects". So we are checking more information.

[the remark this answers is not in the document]
dalg24: Oh right https://github.com/arborx/ArborX/blob/1e28e229b0c5bf0aaae8b908d5f7293cbebe5846/test/tstRay.cpp#L349 is the one that does the most.

masterleinad:
Retest this please.

dalg24 (approved):
Query tree unit tests have a bit too much boiler plate code for my taste but this is good enough.

test/CMakeLists.txt
wjge: not sure of the criterion.

masterleinad:
Retest this please.

aprokop (approved):
LGTM. Please clean or squash the history before merge.

PR #671, opened 2022-04-07
Set a timeout for the nightly CI

masterleinad:
These are the same options we set in .jenkins/continuous.groovy.
`disableConcurrentBuilds(abortPrevious: true)` probably isn't all that relevant here but it should also not hurt.

### DIFF
--- a/.jenkins/nightly.groovy
+++ b/.jenkins/nightly.groovy
@@ -1,4 +1,8 @@
 pipeline {
+    options {
+        disableConcurrentBuilds(abortPrevious: true)
+        timeout(time: 3, unit: 'HOURS')
+    }
     agent none
 
     stages {


.jenkins/nightly.groovy
```diff
@@ -1,4 +1,8 @@
 pipeline {
+    options {
+        disableConcurrentBuilds(abortPrevious: true)
```
dalg24, line 3: Why did you set this option?

masterleinad: I just copied it from the other file and couldn't quite decide to keep it or not. As I said in the initial comment, it probably isn't really relevant here. I'm happy to remove if you prefer.

dalg24: Remove it

aprokop (approved):

PR #677, opened 2022-05-10
[ci skip] Mark png files as binary

aprokop:


masterleinad (approved):
Not quite sure how much we should put into this file but logically this makes sense.

PR #690, opened 2022-06-20
Remove node with two children in favor of node with left child and rope

dalg24:
Following up on https://github.com/arborx/ArborX/pull/672#pullrequestreview-1011018003
Cleaning up and removing `Details::NodeWithTwoChildren`


masterleinad (approved):
Looks good to me. I think it's fine to get rid of the tags altogether. If we want to try something else again, it shouldn't be too hard to come back.

aprokop (approved):

dalg24:
`CUDA-11.0.3-Clang` is hanging.  Not sure if it is worth re-running.

PR #683, opened 2022-05-20
New Raytracing example using PriorityBasedTreeTraversal

masterleinad:
It makes sense to replace the current example since the new one covers more ground, but I could copy some more comments if necessary.
I decided to keep both variants since I find it instructional to see the performance difference.

The `PriorityBasedTreeTraversal` still needs to be moved.

PR #694, opened 2022-07-01
Restrict ViewComparison to non-strided Views

masterleinad:
[arborx/ArborX@`a4e33e3` (#683)](https://github.com/arborx/ArborX/pull/683/commits/a4e33e37045fdbdfd53695a7715f755e8c794b64) exposed that we can't use subviews in `BOOST_TEST` macros since they are strided and, thus, its elements are potentially not contiguous in memory.


dalg24 (approved):
Thanks for catching that bug.
I would like to observe though that here will be another IMO better resolution in the future, that is use Kokkos' random iterators that were added with the STL-like parallel algorithms in release 3.6.

masterleinad:
`RandomAccessIterator` is in namespace `Kokkos::Experimental::Impl`, though. I'm not quite sure if w are happy relying on functionality in an `Impl` namespace.

aprokop (approved):

PR #732, opened 2022-08-31
Add DistributedTree examples

masterleinad:
Supersedes #724 (basically adding to it).

### DIFF
--- a/examples/distributed_tree/distributed_intersects.cpp
+++ b/examples/distributed_tree/distributed_intersects.cpp
@@ -13,12 +13,7 @@
 
 #include <Kokkos_Core.hpp>
 
-#include <cstdarg>
-#include <cstdio>
-#include <iostream>
-#include <random>
-#include <vector>
-
+#include "distributed_setup.hpp"
 #include <mpi.h>
 
 using ExecutionSpace = Kokkos::DefaultExecutionSpace;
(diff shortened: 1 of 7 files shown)


examples/distributed_tree/distributed_intersects.cpp
```diff
@@ -0,0 +1,106 @@
...
+ * the LICENSE file in the top-level directory.                             *
+ *                                                                          *
+ * SPDX-License-Identifier: BSD-3-Clause                                    *
+ ****************************************************************************/
+
+#include <ArborX.hpp>
+
+#include <Kokkos_Core.hpp>
+
+#include <cstdarg>
```
dalg24, line 16: Why did you include that header?

examples/distributed_tree/distributed_intersects.cpp
```diff
@@ -0,0 +1,106 @@
...
+#include <cstdio>
+#include <iostream>
+#include <random>
```
dalg24, line 19: Not used

examples/distributed_tree/distributed_intersects.cpp
```diff
@@ -0,0 +1,106 @@
...
+namespace Example
+{
+template <class Points>
+struct Intersects
+{
+  Points points;
+  float radius;
+};
+template <class Points>
+Intersects(Points const &, int) -> Intersects<Points>;
```
dalg24, line 36:
```suggestion
Intersects(Points, int) -> Intersects<Points>;
```
Wondering whether we should comment that we don't need it from C++20

examples/distributed_tree/distributed_knn.cpp
```diff
@@ -0,0 +1,106 @@
...
+                            static_cast<float>(comm_rank) + .5f,
+                            static_cast<float>(comm_rank) + .5f};
+    std::vector<ArborX::Point> points = {lower_left_corner, center};
+    auto points_device = Kokkos::create_mirror_view_and_copy(
+        MemorySpace{},
+        Kokkos::View<ArborX::Point *, Kokkos::HostSpace,
+                     Kokkos::MemoryUnmanaged>(points.data(), points.size()));
+
+    ExecutionSpace exec;
+    ArborX::DistributedTree<MemorySpace> tree(comm, exec, points_device);
```
dalg24, line 84: This is the same setup 3x
Having a single source file for all scenarios is bad but this code duplication is unfortunate.

PR #759, opened 2022-10-05
Fix the message when error on reading a file

aprokop:


aprokop:
The error in the distributed test is clearly unrelated.

masterleinad (approved):

PR #811, opened 2023-01-05
Use `TeamPolicy` in `Details::expandHalfToFull` to improve performance

dalg24:
In #809 we had opted to merge in the version with `RangePolicy` even though it was commented out with a macro guard in the code I was using to measure performance.
I don't have numbers handy to share but with the `TeamPolicy` version, findHalfNeighborList+expandHalfToFull is 15% faster than the "classic" BVH query implementation, with a `RangePolicy` it is slower than the current status quo.


aprokop (commented):
The patch itself looks fine. I think another minor improvement that should be done is doing
`auto counts = KokkosExt::clone(space, offsets);`
and then doing
`indices(Kokkos::atomic_fetch_inc(&counts(i)) = <blah>;`
inside the parallel_for.

This would
- avoid fetching `offsets` storage
- avoid doing index summation

aprokop (approved):

PR #827, opened 2023-02-08
Do not use `View::Rank`, prefer `View::rank`

dalg24:
`View::Rank` just happens to work.  It is not even documented by Kokkos.  (I am considering whether it should be deprecated.)
Use `View::rank` instead (or View::rank() in later Kokkos versions, that is 4.1+)

### DIFF
--- a/src/details/ArborX_DetailsSortUtils.hpp
+++ b/src/details/ArborX_DetailsSortUtils.hpp
@@ -46,7 +46,7 @@ auto sortObjects(ExecutionSpace const &space, ViewType &view)
 // Helper functions and structs for applyPermutations
 namespace PermuteHelper
 {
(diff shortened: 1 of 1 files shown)


aprokop (commented):
Should the places with `Kokkos::Rank` be replaced too?
```
examples/molecular_dynamics/example_molecular_dynamics.cpp:      Kokkos::MDRangePolicy<Kokkos::Rank<3>, ExecutionSpace>(
examples/raytracing/example_raytracing.cpp:      Kokkos::MDRangePolicy<Kokkos::Rank<3>, ExecutionSpace>(
examples/raytracing/example_raytracing.cpp:        Kokkos::MDRangePolicy<Kokkos::Rank<2>, ExecutionSpace>(
```

aprokop (approved):

dalg24:
`Kokkos::Rank` is something different

PR #837, opened 2023-03-07
Update SYCL docker container avoiding buiding the compiler

masterleinad:
This pull request updates the SYCL compiler to 2023.0.0 which allows us to use Codeplay's CUDA plug-in (https://codeplay.com/solutions/oneapi/for-cuda/) with a public release which improves building the compiler.
Unfortunately, we are seeing some floating point issues with this that partly seem to result from math functions giving different results on host and device even after using `-fp-model=precise`.
The current pull request tries to address them by simply allowing bigger floating point tolerances for the affected tests.

### DIFF
--- a/.jenkins/continuous.groovy
+++ b/.jenkins/continuous.groovy
@@ -412,25 +412,30 @@ pipeline {
                         sh 'rm -rf build && mkdir -p build'
                         dir('build') {
                             sh '''
+                                . /opt/intel/oneapi/setvars.sh --include-intel-llvm && \
                                 cmake \
                                     -D CMAKE_INSTALL_PREFIX=$ARBORX_DIR \
                                     -D CMAKE_BUILD_TYPE=Release \
-                                    -D CMAKE_CXX_COMPILER=clang++ \
+                                    -D CMAKE_CXX_COMPILER=/opt/intel/oneapi/compiler/2023.0.0/linux/bin-llvm/clang++ \
                                     -D CMAKE_CXX_EXTENSIONS=OFF \
-                                    -D CMAKE_CXX_FLAGS="-fsycl-device-code-split=per_kernel -Wpedantic -Wall -Wextra -Wno-unknown-cuda-version" \
-                                    -D CMAKE_PREFIX_PATH="$KOKKOS_DIR;$BOOST_DIR;$BENCHMARK_DIR;$ONE_DPL_DIR" \
+                                    -D CMAKE_CXX_FLAGS="-fp-model=precise -fsycl-device-code-split=per_kernel -Wpedantic -Wall -Wextra -Wno-unknown-cuda-version" \
+                                    -D CMAKE_PREFIX_PATH="$KOKKOS_DIR;$BOOST_DIR;$BENCHMARK_DIR" \
                                     -D ARBORX_ENABLE_MPI=ON \
                                     -D MPIEXEC_PREFLAGS="--allow-run-as-root" \
                                     -D MPIEXEC_MAX_NUMPROCS=4 \
                                     -D ARBORX_ENABLE_TESTS=ON \
                                     -D ARBORX_ENABLE_EXAMPLES=ON \
                                     -D ARBORX_ENABLE_BENCHMARKS=ON \
-                                    -D ARBORX_ENABLE_ONEDPL=ON \
-                                    -D ONEDPL_PAR_BACKEND=serial \
                                 ..
                             '''
-                            sh 'make -j8 VERBOSE=1'
-                            sh 'ctest $CTEST_OPTIONS'
+                            sh '''
+                                . /opt/intel/oneapi/setvars.sh --include-intel-llvm && \
+                                make -j8 VERBOSE=1
+                            '''
+                            sh '''
+                                . /opt/intel/oneapi/setvars.sh --include-intel-llvm && \
+                                ctest $CTEST_OPTIONS
+                            '''
                         }
                     }
                     post {
@@ -444,17 +449,23 @@ pipeline {
                             dir('test_install') {
                                 sh 'cp -r ../examples .'
                                 sh '''
+                                    . /opt/intel/oneapi/setvars.sh --include-intel-llvm && \
                                     cmake \
                                         -D CMAKE_BUILD_TYPE=Release \
-                                        -D CMAKE_CXX_COMPILER=clang++ \
+                                        -D CMAKE_CXX_COMPILER=/opt/intel/oneapi/compiler/2023.0.0/linux/bin-llvm/clang++ \
                                         -D CMAKE_CXX_EXTENSIONS=OFF \
                                         -D CMAKE_CXX_FLAGS="-Wno-unknown-cuda-version" \
-                                        -D CMAKE_PREFIX_PATH="$KOKKOS_DIR;$ARBORX_DIR;$ONE_DPL_DIR" \
-                                        -D ONEDPL_PAR_BACKEND=serial \
+                                        -D CMAKE_PREFIX_PATH="$KOKKOS_DIR;$ARBORX_DIR" \
                                     examples \
                                 '''
-                                sh 'make VERBOSE=1'
-                                sh 'make test'
+                                sh '''
+                                    . /opt/intel/oneapi/setvars.sh --include-intel-llvm && \
+                                    make VERBOSE=1
+                                '''
+                                sh '''
+                                    . /opt/intel/oneapi/setvars.sh --include-intel-llvm && \
+                                    make test
+                                '''
                             }
                         }
                     }
--- a/.jenkins/continuous.groovy
+++ b/.jenkins/continuous.groovy
@@ -416,7 +416,7 @@ pipeline {
                                 cmake \
                                     -D CMAKE_INSTALL_PREFIX=$ARBORX_DIR \
                                     -D CMAKE_BUILD_TYPE=Release \
-                                    -D CMAKE_CXX_COMPILER=/opt/intel/oneapi/compiler/2023.0.0/linux/bin-llvm/clang++ \
+                                    -D CMAKE_CXX_COMPILER=${DPCPP} \
                                     -D CMAKE_CXX_EXTENSIONS=OFF \
                                     -D CMAKE_CXX_FLAGS="-fp-model=precise -fsycl-device-code-split=per_kernel -Wpedantic -Wall -Wextra -Wno-unknown-cuda-version" \
                                     -D CMAKE_PREFIX_PATH="$KOKKOS_DIR;$BOOST_DIR;$BENCHMARK_DIR" \
@@ -453,7 +453,7 @@ pipeline {
                                     . /opt/intel/oneapi/setvars.sh --include-intel-llvm && \
                                     cmake \
                                         -D CMAKE_BUILD_TYPE=Release \
-                                        -D CMAKE_CXX_COMPILER=/opt/intel/oneapi/compiler/2023.0.0/linux/bin-llvm/clang++ \
+                                        -D CMAKE_CXX_COMPILER=${DPCPP} \
                                         -D CMAKE_CXX_EXTENSIONS=OFF \
                                         -D CMAKE_CXX_FLAGS="-Wno-unknown-cuda-version" \
                                         -D CMAKE_PREFIX_PATH="$KOKKOS_DIR;$ARBORX_DIR" \
--- a/.jenkins/continuous.groovy
+++ b/.jenkins/continuous.groovy
@@ -426,6 +426,7 @@ pipeline {
                                     -D ARBORX_ENABLE_TESTS=ON \
                                     -D ARBORX_ENABLE_EXAMPLES=ON \
                                     -D ARBORX_ENABLE_BENCHMARKS=ON \
+                                    -D ARBORX_ENABLE_ONEDPL=ON \
                                 ..
                             '''
                             sh '''
--- a/docker/Dockerfile.sycl
+++ b/docker/Dockerfile.sycl
@@ -53,6 +53,7 @@ RUN wget https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCT
     apt-get install -y intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic-${DPCPP_VERSION} && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
+ENV DPCPP=/opt/intel/oneapi/compiler/${DPCPP_VERSION}/linux/bin-llvm/clang++
 
 # Install Codeplay's oneAPI for NVIDIA GPUs, see
 # https://developer.codeplay.com/products/oneapi/nvidia/2023.0.0/guides/get-started-guide-nvidia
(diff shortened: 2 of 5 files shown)


masterleinad:
Retest this please.

masterleinad:
Retest this please.

masterleinad:
Retest this please.

.jenkins/continuous.groovy
```diff
@@ -412,12 +51,13 @@ pipeline {
...
                             sh '''
+                                . /opt/intel/oneapi/setvars.sh --include-intel-llvm && \
                                 cmake \
                                     -D CMAKE_INSTALL_PREFIX=$ARBORX_DIR \
                                     -D CMAKE_BUILD_TYPE=Release \
-                                    -D CMAKE_CXX_COMPILER=clang++ \
+                                    -D CMAKE_CXX_COMPILER=/opt/intel/oneapi/compiler/2023.0.0/linux/bin-llvm/clang++ \
                                     -D CMAKE_CXX_EXTENSIONS=OFF \
-                                    -D CMAKE_CXX_FLAGS="-fsycl-device-code-split=per_kernel -Wpedantic -Wall -Wextra -Wno-unknown-cuda-version" \
+                                    -D CMAKE_CXX_FLAGS="-fp-model=precise -fsycl-device-code-split=per_kernel -Wpedantic -Wall -Wextra -Wno-unknown-cuda-version" \
```
aprokop, line 60: Can we try it with `-fp-model source`? Few online sources seem to indicate this is required for IEEE compliance. It also implies `precise`.

masterleinad: That option doesn't exist, see https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2023-1/fp-model-fp.html. I also tried `strict` with the same test failures.

aprokop: Ah, I was looking at the Intel compiler, not DPC++

masterleinad:
Only HIP CI is timing out. Everything else is passing.

.jenkins/continuous.groovy
```diff
@@ -412,25 +412,30 @@ pipeline {
...
+                                    -D CMAKE_PREFIX_PATH="$KOKKOS_DIR;$BOOST_DIR;$BENCHMARK_DIR" \
                                     -D ARBORX_ENABLE_MPI=ON \
                                     -D MPIEXEC_PREFLAGS="--allow-run-as-root" \
                                     -D MPIEXEC_MAX_NUMPROCS=4 \
                                     -D ARBORX_ENABLE_TESTS=ON \
                                     -D ARBORX_ENABLE_EXAMPLES=ON \
                                     -D ARBORX_ENABLE_BENCHMARKS=ON \
-                                    -D ARBORX_ENABLE_ONEDPL=ON \
-                                    -D ONEDPL_PAR_BACKEND=serial \
```
dalg24, line 429: We cannot use ONEDPL with that image?

docker/Dockerfile.sycl
```diff
@@ -43,63 +43,45 @@ RUN CMAKE_KEY=2D2CEF1034921684 && \
...
-    gpg --verify ${BOOST_ARCHIVE}.json.asc ${BOOST_ARCHIVE}.json && \
-    gpg --verify ${BOOST_ARCHIVE}.asc ${BOOST_ARCHIVE} && \
     cat ${BOOST_ARCHIVE}.json | jq -r '. | .sha256 + "  " + .file' | sha256sum --check && \
     mkdir -p boost && \
     tar -xf ${BOOST_ARCHIVE} -C boost --strip-components=1 && \
     cd boost && \
-    CXXFLAGS="-w" ./bootstrap.sh \
-        --prefix=${BOOST_DIR} \
-        && \
+    sed -i "s/-pch-create/-Xclang -emit-pch -o/g" tools/build/src/tools/intel-linux.jam && \
```
aprokop, line 79: Can you please elaborate on this? This is magic to me.

docker/Dockerfile.sycl
```diff
@@ -118,17 +100,19 @@ RUN SCRATCH_DIR=/scratch && mkdir -p ${SCRATCH_DIR} && cd ${SCRATCH_DIR} && \
     rm -rf ${SCRATCH_DIR}
 
 # Install Kokkos
-ARG KOKKOS_VERSION=3.7.01
+# Kokkos 4.0.99 using in-order queues for SYCL+Cuda
```
aprokop, line 103: Why do we need in-order queues?

docker/Dockerfile.sycl
```diff
@@ -118,17 +100,19 @@ RUN SCRATCH_DIR=/scratch && mkdir -p ${SCRATCH_DIR} && cd ${SCRATCH_DIR} && \
...
+    KOKKOS_URL=https://github.com/kokkos/kokkos/archive/${KOKKOS_VERSION}.tar.gz && \
     KOKKOS_ARCHIVE=kokkos-${KOKKOS_HASH}.tar.gz && \
     SCRATCH_DIR=/scratch && mkdir -p ${SCRATCH_DIR} && cd ${SCRATCH_DIR} && \
     wget --quiet ${KOKKOS_URL} --output-document=${KOKKOS_ARCHIVE} && \
     mkdir -p kokkos && \
     tar -xf ${KOKKOS_ARCHIVE} -C kokkos --strip-components=1 && \
     cd kokkos && \
     mkdir -p build && cd build && \
-    cmake -D CMAKE_BUILD_TYPE=Release -D CMAKE_INSTALL_PREFIX=${KOKKOS_DIR} -D CMAKE_CXX_COMPILER=clang++ ${KOKKOS_OPTIONS} .. && \
+    cmake -D CMAKE_BUILD_TYPE=Release -D CMAKE_INSTALL_PREFIX=${KOKKOS_DIR} -D CMAKE_CXX_COMPILER=/opt/intel/oneapi/compiler/2023.0.0/linux/bin-llvm/clang++ ${KOKKOS_OPTIONS} .. && \
```
aprokop, line 116: Please use some variable here, so that once you change the Intel compiler version above, you don't need to change it here.

test/tstQueryTreeCallbacks.cpp
```diff
@@ -75,18 +75,22 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(callback_spatial_predicate, TreeTypeTraits,
...
+      "values_device", n);
+  Kokkos::parallel_for(
+      Kokkos::RangePolicy<ExecutionSpace>(0, n), KOKKOS_LAMBDA(int i) {
+        ArborX::Point const origin = {{0., 0., 0.}};
+        values_device(i) = {i, ArborX::Details::distance(points(i), origin)};
+      });
+  std::vector<Kokkos::pair<int, float>> values(n);
+  Kokkos::deep_copy(Kokkos::View<Kokkos::pair<int, float> *, Kokkos::HostSpace>(
+                        values.data(), n),
+                    values_device);
```
aprokop, line 93: How about encapsulating this into a helper function? Would also make the device side initiation only for SYCL, and use the standard one for everything else.

Here and the other 3 spots below.

.jenkins/continuous.groovy
```diff
@@ -412,25 +412,30 @@ pipeline {
                         sh 'rm -rf build && mkdir -p build'
                         dir('build') {
                             sh '''
+                                . /opt/intel/oneapi/setvars.sh --include-intel-llvm && \
                                 cmake \
                                     -D CMAKE_INSTALL_PREFIX=$ARBORX_DIR \
                                     -D CMAKE_BUILD_TYPE=Release \
-                                    -D CMAKE_CXX_COMPILER=clang++ \
+                                    -D CMAKE_CXX_COMPILER=/opt/intel/oneapi/compiler/2023.0.0/linux/bin-llvm/clang++ \
```
aprokop, line 419: Maybe create a link somewhere so that we don't have to spell version?

aprokop (commented):
This was a really tough job to make this work. I have a few comments about clarifying a few things.

.jenkins/continuous.groovy
masterleinad: Yes, we can. We make sure that the compiler already has the appropriate includes and doesn't need to find a  `CMake` package. We would error out if `oneDPL` could not be found (since it's enabled by default).

.jenkins/continuous.groovy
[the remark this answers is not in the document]
masterleinad: That was my experience when I tried it first. I'm happy to add a commit to check if that's still the case.

docker/Dockerfile.sycl
```diff
@@ -43,63 +43,45 @@ RUN CMAKE_KEY=2D2CEF1034921684 && \
...
-    mkdir -p ${SYCL_DIR} && \
-    mv ${SCRATCH_DIR}/llvm/build/install/* ${SYCL_DIR} && \
-    echo "${SYCL_DIR}/lib" > /etc/ld.so.conf.d/sycl.conf && ldconfig && \
-    rm -rf ${SCRATCH_DIR}
-ENV PATH=${SYCL_DIR}/bin:$PATH
+RUN wget https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB && \
+    apt-key add GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB && \
+    echo "deb https://apt.repos.intel.com/oneapi all main" | tee /etc/apt/sources.list.d/oneAPI.list && \
+    apt-get update -o Dir::Etc::sourcelist="sources.list.d/oneAPI.list" -o APT::Get::List-Cleanup="0" && \
+    apt-get install -y intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic-2023.0.0 && \
```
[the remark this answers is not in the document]
masterleinad: https://www.intel.com/content/www/us/en/docs/vtune-profiler/installation-guide/2023-1/package-managers.html

docker/Dockerfile.sycl
```diff
@@ -43,63 +43,45 @@ RUN CMAKE_KEY=2D2CEF1034921684 && \
...
-    mkdir onedpl && \
-    tar -xf ${ONE_DPL_ARCHIVE} -C onedpl --strip-components=1 && cd onedpl && \
-    echo 'install(CODE "set(OUTPUT_DIR \"${CMAKE_INSTALL_PREFIX}/lib/cmake/oneDPL\")")' >> CMakeLists.txt && \
-    echo 'install(SCRIPT cmake/scripts/generate_config.cmake)' >> CMakeLists.txt && \
-    echo 'install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/include DESTINATION linux)' >> CMakeLists.txt && \
-    mkdir build && cd build && \
-    cmake -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_CXX_FLAGS="-w" -DCMAKE_INSTALL_PREFIX=${ONE_DPL_DIR} -DCMAKE_SKIP_INSTALL_ALL_DEPENDENCY=TRUE -DONEDPL_BACKEND="dpcpp_only" .. && \
-    make -j${NPROCS} install && \
-    rm -rf ${SCRATCH_DIR}
+RUN wget https://cloud.cees.ornl.gov/download/oneapi-for-nvidia-gpus-2023.0.0-linux.sh && \
```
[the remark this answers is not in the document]
masterleinad: We want to use Codeplay's NVIDIA plugin and the most convenient way was to upload it into the cloud (which we had already done for `Kokkos` testing).

docker/Dockerfile.sycl
```diff
@@ -43,63 +43,45 @@ RUN CMAKE_KEY=2D2CEF1034921684 && \
...
+    chmod +x oneapi-for-nvidia-gpus-2023.0.0-linux.sh && \
+    ./oneapi-for-nvidia-gpus-2023.0.0-linux.sh -y && \
+    rm oneapi-for-nvidia-gpus-2023.0.0-linux.sh
+
+RUN wget https://registrationcenter-download.intel.com/akdlm/irc_nas/19133/l_oneDPL_p_2022.0.0.25335.sh &&\
```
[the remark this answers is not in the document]
masterleinad: We install oneDPL.

docker/Dockerfile.sycl
```diff
@@ -43,63 +43,45 @@ RUN CMAKE_KEY=2D2CEF1034921684 && \
...
+    chmod +x ./l_oneDPL_p_2022.0.0.25335.sh && \
+    ./l_oneDPL_p_2022.0.0.25335.sh -a -s --eula accept && \
+    rm l_oneDPL_p_2022.0.0.25335.sh
 
 # Install Boost
 ENV BOOST_DIR=/opt/boost
-RUN BOOST_VERSION=1.72.0 && \
+RUN . /opt/intel/oneapi/setvars.sh --include-intel-llvm && \
+    BOOST_VERSION=1.81.0 && \
```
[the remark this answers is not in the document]
masterleinad: Prior versions don't support the compiler and we still need fixes, see https://www.intel.com/content/www/us/en/developer/articles/technical/building-boost-with-oneapi.html.

.jenkins/continuous.groovy
dalg24: Is that something you intend to revisit later?

docker/Dockerfile.sycl
```diff
@@ -43,63 +43,45 @@ RUN CMAKE_KEY=2D2CEF1034921684 && \
...
-    make -j${NPROCS} install && \
-    rm -rf ${SCRATCH_DIR}
+RUN wget https://cloud.cees.ornl.gov/download/oneapi-for-nvidia-gpus-2023.0.0-linux.sh && \
+    chmod +x oneapi-for-nvidia-gpus-2023.0.0-linux.sh && \
+    ./oneapi-for-nvidia-gpus-2023.0.0-linux.sh -y && \
+    rm oneapi-for-nvidia-gpus-2023.0.0-linux.sh
+
+RUN wget https://registrationcenter-download.intel.com/akdlm/irc_nas/19133/l_oneDPL_p_2022.0.0.25335.sh &&\
+    chmod +x ./l_oneDPL_p_2022.0.0.25335.sh && \
+    ./l_oneDPL_p_2022.0.0.25335.sh -a -s --eula accept && \
```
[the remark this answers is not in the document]
masterleinad: I have never seen the CUDA version to matter, and the oneDPL version was simply the newest version available which I hoped would work well.

.jenkins/continuous.groovy
```diff
@@ -412,25 +412,30 @@ pipeline {
                         sh 'rm -rf build && mkdir -p build'
                         dir('build') {
                             sh '''
+                                . /opt/intel/oneapi/setvars.sh --include-intel-llvm && \
```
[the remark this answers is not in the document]
dalg24: It would be better to tackle in the docker image and make sure the script is sourced when you start a bash session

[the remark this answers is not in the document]
masterleinad: No, AFAICT there is no good solution.

.jenkins/continuous.groovy
masterleinad: No, this is using oneDPL.

dalg24: How so?  Presumably the macro `ARBORX_ENABLE_ONEDPL` won't be defined.

masterleinad: I don't understand, `ARBORX_ENABLE_ONEDPL` is by default `ON` when `Kokkos_ENABLE_SYCL` is defined.
It seems it's easier to just be explicit and add it back.

dalg24: My bad I misread the cmake_dependent_option.  Sorry about the noise.

masterleinad:
Retest this please.

test/tstQueryTreeCallbacks.cpp
aprokop: I'm not opposed to having a single path. What about a separate function, though? It seems that we could reduce the 4x duplication of the code with it.

aprokop (approved):

aprokop:
SYCL build passed, couple builds timed out.

PR #887, opened 2023-06-05
Remove old unused code in BatchedQueries

aprokop:


dalg24 (approved):
Fine by me.  If I recall correctly, some of this functionality was never actually used but kept because we thought we might need it at some point.

PR #901, opened 2023-06-19
Fix include files for ArborX_DetailsKokkosExtBitManipulation.hpp and Kokkos >= 4.1.00

masterleinad:
Apparently, we only test for Kokkos >= 4.1.00 in our nightlies which don't build the tests but only the benchmarks.


aprokop (approved):

dalg24 (commented):
Just realized I had forgotten to submit.
Was wondering why you merged without commenting :/

PR #923, opened 2023-08-08
Suppress -Wsycl-target warnings for SYCL CI

masterleinad:
We are seeing
```
clang++: warning: linked binaries do not contain expected 'nvptx64-nvidia-cuda' target; found targets: 'nvptx64-nvidia-cuda-sm_70' [-Wsycl-target]
```
a bunch in the CI. Supposedly, this was fixed by https://github.com/intel/llvm/pull/7434. It shouldn't matter much anyway (and if so it would be more of a `Kokkos` issue than an `ArborX` issue) and there is no reason to fail the CI because of this warning.

### DIFF
--- a/.jenkins/continuous.groovy
+++ b/.jenkins/continuous.groovy
@@ -421,7 +421,7 @@ pipeline {
                                     -D CMAKE_BUILD_TYPE=Release \
                                     -D CMAKE_CXX_COMPILER=${DPCPP} \
                                     -D CMAKE_CXX_EXTENSIONS=OFF \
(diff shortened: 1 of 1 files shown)


masterleinad:
> Could this warning be fixed by updating the SYCL container?
It seems like it should have been fixed in 2022-12, but we are using 2023-01 in our CI. Does it mean it's not?

I expect it to be a result of the combination of the `oneAPI` compiler and the `Codeplay` plugin we are using. It might be fixed in newer versions but I don't see this warning enough of a reason to upgrade.

aprokop (approved):
LGTM

PR #929, opened 2023-08-15
Replace ARBORX_CLASS_LAMBDA with KOKKOS_CLASS_LAMBDA

masterleinad:
We require `Kokkos 4.0.00` by now so we don't have to roll our own anymore.


aprokop (approved):
LGTM

PR #935, opened 2023-08-24
Disable deprecation when building Kokkos in nightly regression testing

masterleinad:
This should suppress the warnings in https://cloud.cees.ornl.gov/jenkins-ci/job/ArborX-nightly/.

### DIFF
--- a/.jenkins/nightly.groovy
+++ b/.jenkins/nightly.groovy
@@ -26,7 +26,7 @@ pipeline {
                         dir('source-kokkos') {
                             sh 'git rev-parse --short HEAD'
                         }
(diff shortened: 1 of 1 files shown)


dalg24 (changes requested):
Why would we do this.  We definitely want to know if we use something deprecated.

aprokop (commented):
I don't understand the motivation for it. If there are deprecation warnings coming from the Kokkos that ArborX shows, shouldn't they be fixed in Kokkos? Once they are in Kokkos `develop`, they would automatically disappear from our nightlies.

PR #945, opened 2023-09-06
Fix bug argument counter must be passed by reference

dalg24:
Passing an argument recognized by Kokkos to a test executable yields a segfault.
The issue is that Kokkos decrement the argument counter and shift the arguments but Boost.Test gets the original count and dereferences a null pointer.


aprokop (commented):
Please update the copyright year.

Why has this never failed for us?

dalg24:
> Please update the copyright year.

Done

> Why has this never failed for us?

Because no-one tried before

aprokop (approved):

dalg24:
HIP build timed out.  I suggest to ignore.  I don't think it is worth re-running the whole CI.

PR #962, opened 2023-10-23
Implement APIv2 BruteForce

aprokop:
Will need to be rebased on #915.

PR #966, opened 2023-10-31
Make the BruteForce interface uniform with BoundingVolumeHierarchy with respect to TraversalPolicy

masterleinad:
Addressing https://github.com/arborx/ArborX/pull/962#discussion_r1376627553. We can't really check the values here since we would like to pass values that are only relevant for BVH.

### DIFF
--- a/src/ArborX_BruteForce.hpp
+++ b/src/ArborX_BruteForce.hpp
@@ -58,10 +58,11 @@ public:
   KOKKOS_FUNCTION
   bounding_volume_type bounds() const noexcept { return _bounds; }
 
-  template <typename ExecutionSpace, typename Predicates, typename Callback,
-            typename Ignore = int>
+  template <typename ExecutionSpace, typename Predicates, typename Callback>
   void query(ExecutionSpace const &space, Predicates const &predicates,
-             Callback const &callback, Ignore = Ignore()) const;
+             Callback const &callback,
+             Experimental::TraversalPolicy const &policy =
+                 Experimental::TraversalPolicy()) const;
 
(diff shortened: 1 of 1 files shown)


src/ArborX_BruteForce.hpp
```diff
@@ -58,10 +58,11 @@ class BasicBruteForce
...
   bounding_volume_type bounds() const noexcept { return _bounds; }
 
-  template <typename ExecutionSpace, typename Predicates, typename Callback,
-            typename Ignore = int>
+  template <typename ExecutionSpace, typename Predicates, typename Callback>
   void query(ExecutionSpace const &space, Predicates const &predicates,
-             Callback const &callback, Ignore = Ignore()) const;
+             Callback const &callback,
+             Experimental::TraversalPolicy const &policy =
+                 Experimental::TraversalPolicy()) const;
```
dalg24, line 65: `Experimental::TraversalPolicy` makes absolutely no sense for `BruteForce`.  This was literally the point of that `Ignore` parameter.

dalg24 (changes requested):

dalg24:
What issue is that resolving?  The `Ignore` template parameter is precisely there for compatibility.

src/ArborX_BruteForce.hpp
masterleinad: It just doesn't make much sense to me to have a user-exposed template parameter that allows passing in arbitrary types that are entirely ignored.

masterleinad:
We hope to refactor the interface for BVH that might get rid of `TraversalPolicy` anyway.

PR #988, opened 2023-12-20
Drop check_valid_geometry_traits test from evaluatePolynomialBasis

masterleinad:
This should avoid some changes in https://github.com/arborx/ArborX/pull/946, like https://github.com/arborx/ArborX/pull/946/files#diff-a2fce1f8ea3d6b4f0053b59757e3a38d63072a53b10f24a2b731821423a95cb8R93. We already test
```
static_assert(GeometryTraits::is_point<Point>::value,
                "point must be a point");
```
and we use traits without further checks in other places already.


aprokop (approved):

mrlag31:
What can also be done is removing the `KOKKOS_FUNCTION constexpr` of `check_valid_geometry_traits`, as this call was the reason I added those

PR #1001, opened 2023-12-28
Modernize and fix callback detection

dalg24:
Refactor/ modernize callback introspection facility.  My initial goal was to get rid of `Details::AccessTraitsHelper` but I decided to split that up.

Fix bug introduced the refactor in #939 that makes that we were not catching errors like https://github.com/arborx/ArborX/blob/1dc256dff44ae10847175258dc43c750d1142240/test/tstCompileOnlyCallbacks.cpp#L141-L142))


aprokop (approved):
LGTM

PR #1019, opened 2024-01-09
Remove GitLab performance regression CI

masterleinad:
It seems that we won't be able to get the performance regression CI back to work in the same way we used it before ECP ended.


aprokop (commented):
Do we want to remove it? I disabled the actions, so it should not trigger. I don't see a pressing need for removal, though I understand the desire.

masterleinad:
> Do we want to remove it? I disabled the actions, so it should not trigger. I don't see a pressing need for removal, though I understand the desire.

It's just dead code at this point and having it makes the impression that we are using it while we don't. I also still get notifications from forks, see, e.g. https://github.com/dalg24/ArborX/actions/runs/7475756660.

aprokop (approved):

PR #1021, opened 2024-01-10
An attempt to cleanup symbols for values and predicates

dalg24:
The recent refactors makes us generate symbol names that are getting really nasty.  This is an attempt to "collapse" nested template template classes when possible to clean things up.  My fancy debugger does not let me copy/paste symbol name and I am too lazy/tired to type it down before and after but I think you will get the picture.

This is meant as a basis for discussion.  I went for minimal changes, I am not suggesting that aliasing `AccessValues` to some implementation class is a good design.

I implemented the following contractions
`AccessValues<AccessValues<...>> -> AccessValues<...>`
`AccessValues<Kokkos::View<...>> -> Kokkos::View<...>`
`AccessValues<PermutedData<...>> -> PermutedData<...>`
`AccessValues<LegacyValues<...>> -> LegacyValues<...>`

### DIFF
--- a/src/details/ArborX_AccessTraits.hpp
+++ b/src/details/ArborX_AccessTraits.hpp
@@ -190,14 +190,14 @@ void check_valid_access_traits(PrimitivesTag, Primitives const &,
 }
 
 template <typename Values, typename Tag>
-class AccessValues
+class AccessValuesI
 {
 private:
   using Access = AccessTraits<Values, Tag>;
   Values _values;
 
 public:
-  explicit AccessValues(Values values)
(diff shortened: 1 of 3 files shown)


dalg24:
We agreed to merge this minimal change set and deal with the call sites in a follow-up

dalg24:
`AccessValuesI` stands for "Implementation" and we went for the shorthand notation because `AccessValues` is an alias so the underlying type name gets printed and we wanted the keep the symbol name short.  We will rename later when refactoring the call sites.

aprokop (approved):

PR #1044, opened 2024-03-14
Enable [ci skip] in Jenkins continuous pipeline

dalg24:
(Re)enable skipping builds if the latest commit message says "[ci skip]"
It used to be in place before our server went down in last November.  Being able to skip a build and save resources is worthwhile.

See [plugin doc](https://plugins.jenkins.io/scmskip/#plugin-content-enable-on-pipeline)


dalg24:
Did NOT work...

PR #1048, opened 2024-04-01
Switch CI from MI100 to MI210

Rombur:
The MI210 machines seem more stable than the MI100 ones. Let's just use them instead.


masterleinad (approved):

aprokop (approved):
LGTM. HIP build passed.

PR #1057, opened 2024-04-09
[ci skip] Add examples README

aprokop:


masterleinad (commented):
What about a table instead for better alignment of the description?

dalg24 (approved):
It is an improvement over the status quo.

PR #1074, opened 2024-04-19
Move `OrderedSpatialPredicateTag` where it belongs that is namespace `Details::`

dalg24:
The predicate tags really are implementation details and do not belong to the `Experimental::` namespace.


aprokop (approved):
Sure. What was the reason you decided to look at that?

dalg24:
Looking into something else (refactoring of AccessValues to check validity
at construction when adapting user input)

PR #1154, opened 2024-09-22
Make hyper triangle default

aprokop:


aprokop:
Ignoring unrelated failures.

PR #1210, opened 2025-02-04
Decrease number of procs used to build ArborX in the nightly

Rombur:
The nightly randomly fails on GCC 14. Decreasing the number of processors to see if that fixes the issue.


aprokop (approved):
Sure.

PR #1265, opened 2025-06-13
Bump johnwason/vcpkg-action from 6 to 7 and fix Windows build

aprokop:


aprokop:
Windows build passed, that's all we care about.

PR #1305, opened 2025-10-30
Fix few instances of the wrong const qualifier order

aprokop:
I don't understand why `clang-format` does nothing with those.


dalg24 (approved):
Fun stuff

PR #1328, opened 2026-01-15
Add intersects(sphere, sphere) and distance(sphere, sphere)

aprokop:


masterleinad (approved):

aprokop:
HIP ran out of storage, ignoring.

PR #1330, opened 2026-01-18
Enable reverse dispatch for intersects and distance algos

dalg24:
Synthesize the reverse dispatch `algo(geom2, geom1)` and only require one specialization.

An attempt to resolve the issue Andrey reported about some combination of algo and geometries that we forgot to implement.


dalg24:
I had not seen your PR until now :/ 
I think the advantage with the reverse dispatch helper is that we don't repeat the logic for each algorithm.

PR #1356, opened 2026-05-11
Distributed DBSCAN timers and cleanup

aprokop:
- Add an additional details function to reduce the amount of code in the distributed `dbscan` function
- Add profiling regions for each step of the distributed algorithm
- Actually hook the timers for algorithm steps and print them out in verbose
- Update `Timer` used in `TimeMonitor` to be able to explicitly set the time

### DIFF
--- a/benchmarks/cluster/CMakeLists.txt
+++ b/benchmarks/cluster/CMakeLists.txt
@@ -23,7 +23,7 @@ target_link_libraries(ArborX_Benchmark_HDBSCAN.exe ArborX::ArborX Boost::program
 add_test(NAME ArborX_Benchmark_MST COMMAND ArborX_Benchmark_MST.exe --ascii --filename=${input_file})
 
 if (ARBORX_ENABLE_MPI)
-  set(ARBORX_BENCHMARK_UTILS_INCLUDE_DIR ${CMAKE_SOURCE_DIR}/benchmarks/utils)
+  set(ARBORX_BENCHMARK_UTILS_INCLUDE_DIR ${PROJECT_SOURCE_DIR}/benchmarks/utils)
   add_executable(ArborX_Benchmark_DistributedDBSCAN.exe distributed_dbscan.cpp)
   target_include_directories(ArborX_Benchmark_DistributedDBSCAN.exe PRIVATE ${CMAKE_CURRENT_SOURCE_DIR} ${ARBORX_BENCHMARK_UTILS_INCLUDE_DIR})
   target_link_libraries(ArborX_Benchmark_DistributedDBSCAN.exe ArborX::ArborX Boost::program_options cluster_benchmark_helpers)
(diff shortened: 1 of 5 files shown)


benchmarks/utils/ArborXBenchmark_TimeMonitor.hpp
```diff
@@ -66,6 +66,16 @@ class TimeMonitor
...
+    // Explicitly set the duration. This is useful for copying Kokkos timers
+    // we get from hooking Kokkos profiling tools into our own TimeMonitor
+    // timers. In that case, we don't have the start/stop semantics, but we
+    // just want to set the duration of the timer to the value we get from
+    // Kokkos timer.
+    void set(double duration)
+    {
+      assert(!_started);
+      _entry.second = duration;
+    }
```
aprokop, line 78: @dalg24 Thoughts on this?

benchmarks/cluster/distributed_dbscan.cpp
```diff
@@ -128,7 +147,7 @@ int main(int argc, char *argv[])
     --argc;
   }
 
-  Kokkos::ScopeGuard guard(argc, argv);
+  Kokkos::initialize(argc, argv);
```
aprokop, line 150: This change was required to be able to run space-time-stack with MPI enabled profiling.

masterleinad: Why is that? What's the problem you were running into?

aprokop:
```
*** The MPI_Comm_rank() function was called after MPI_FINALIZE was invoked.
*** This is disallowed by the MPI standard.
*** Your MPI job will now abort.
```
So space-time-stack tries to use `MPI_Comm_rank` after MPI is finalized, when `ScopeGuard` is destroyed I assume.

masterleinad: Ah, `Kokkos::finalize` calls the destructor for the space-time-stack objects and needs MPI to be alive.

benchmarks/cluster/CMakeLists.txt
```diff
@@ -23,8 +23,9 @@ target_link_libraries(ArborX_Benchmark_HDBSCAN.exe ArborX::ArborX Boost::program
 add_test(NAME ArborX_Benchmark_MST COMMAND ArborX_Benchmark_MST.exe --ascii --filename=${input_file})
 
 if (ARBORX_ENABLE_MPI)
+  set(ARBORX_BENCHMARK_UTILS_INCLUDE_DIR ${CMAKE_SOURCE_DIR}/benchmarks/utils)
```
dalg24:
```suggestion
  set(ARBORX_BENCHMARK_UTILS_INCLUDE_DIR ${PROJECT_SOURCE_DIR}/benchmarks/utils)
```

benchmarks/utils/ArborXBenchmark_TimeMonitor.hpp
dalg24: It sounds weird.
I don't get why we would want to do that with the current context
I am not saying it is wrong, I just don't follow.

aprokop: What I want is to use hooks with profiling regions and then get MPI statistics out of it. So, I need to transfer the timers we got from hooks into machinery of TimeMonitor so that it can compute global stats (min, mean, max).

PR #1357, opened 2026-05-12
Use PROJECT_SOURCE_DIR instead of CMAKE_SOURCE_DIR

aprokop:


dalg24 (approved):

aprokop:
One test is failing, but I don't see how this change is relevant at all. So am going to ignore it.

PR #1360, opened 2026-05-20
Clean up triangle-sphere intersect test

aprokop:

PR #1361, opened 2026-05-22
Destroy execution space before Kokkos finalize in distributed DBSCAN benchmark

aprokop:


Rombur (approved):

masterleinad (approved):

masterleinad (commented):
Alternatively, you could also just create a temporary in te function call.

On a separate note, it seems only the last value written to `succes` is ever checked.

aprokop:
> Alternatively, you could also just create a temporary in the function call.

I thought about it and opted to go through the minimum number of changes.

> On a separate note, it seems only the last value written to `success` is ever checked.

`success` is only updated once.

masterleinad:
> I thought about it and opted to go through the minimum number of changes.

Replacing with temporaries would be a 4 line change instead of 33. 🙂

aprokop:
SYCL failure will be fixed in #1360.

PR #1385, opened 2026-08-18
Improve scratch size allocation mechanism for MLS

masterleinad:
This pull request fixes the MovingLeastSquares unit tests when running with HIP by using scratch level 1. With https://github.com/kokkos/kokkos/pull/9339 we could also use scratch level 0 with the tests.

Generally, this pull request automatically switches to scratch level 1 when there isn't enough memory available for scratch level 0.

### DIFF
--- a/src/interpolation/detail/ArborX_InterpDetailsMovingLeastSquaresCoefficients.hpp
+++ b/src/interpolation/detail/ArborX_InterpDetailsMovingLeastSquaresCoefficients.hpp
@@ -67,26 +67,20 @@ public:
       , _num_targets(target_access.size())
       , _num_neighbors(source_points.extent_int(1))
   {
-// FIXME_HIP The HIP backend is limited by the small level 0 scratch space since
-// Kokkos through at least version 5.2 requires at least enough scratch memory
-// to run with a workgroup size of 64.
-#ifdef KOKKOS_ENABLE_HIP
     if (perTargetMem() >
         static_cast<std::size_t>(
             Kokkos::TeamPolicy<ExecutionSpace>::scratch_size_max(1)))
       Kokkos::abort("Can't allocate enough scratch space!");
+
+#ifdef KOKKOS_ENABLE_HIP
+    // FIXME_HIP The HIP backend is limited by the small level 0 scratch space
+    // since Kokkos through at least version 5.2 requires at least enough
+    // scratch memory to run with a workgroup size of 64.
     _scratch_level = 1;
 #else
-    if (perTargetMem() <=
-        static_cast<std::size_t>(
-            Kokkos::TeamPolicy<ExecutionSpace>::scratch_size_max(0)))
-      _scratch_level = 0;
-    else if (perTargetMem() <=
-             static_cast<std::size_t>(
-                 Kokkos::TeamPolicy<ExecutionSpace>::scratch_size_max(1)))
-      _scratch_level = 1;
-    else
-      Kokkos::abort("Can't allocate enough scratch space!");
+    std::size_t scratch_0_max_size =
+        Kokkos::TeamPolicy<ExecutionSpace>::scratch_size_max(0);
+    _scratch_level = (perTargetMem() > scratch_0_max_size);
 #endif
   }
 
--- a/src/interpolation/detail/ArborX_InterpDetailsMovingLeastSquaresCoefficients.hpp
+++ b/src/interpolation/detail/ArborX_InterpDetailsMovingLeastSquaresCoefficients.hpp
@@ -71,7 +71,7 @@ public:
   template <typename TeamMember>
   KOKKOS_FUNCTION void operator()(TeamMember const &member) const
   {
-    auto const &scratch = member.thread_scratch(0);
+    auto const &scratch = member.thread_scratch(_scratch_level);
 
     int target = member.league_rank() * member.team_size() + member.team_rank();
     if (target >= _num_targets)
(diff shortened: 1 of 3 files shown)


src/interpolation/detail/ArborX_InterpDetailsMovingLeastSquaresCoefficients.hpp
```diff
@@ -199,19 +199,34 @@ class MovingLeastSquaresCoefficientsKernel
...
+      _scratch_level = 1;
+    else
+      Kokkos::abort("Can't allocate enough scratch space!");
+#endif
+
+    Kokkos::TeamPolicy policy(space, 1, Kokkos::AUTO);
+    policy.set_scratch_size(_scratch_level, Kokkos::PerThread(perTargetMem()));
     int team_size =
-        dummy_policy.team_size_recommended(*this, Kokkos::ParallelForTag{});
-    if (team_size != 0)
```
masterleinad, line 207: This case can't occur (anymore). If we can't find a suitable workgroup size, `Kokkos` errors out.

src/interpolation/detail/ArborX_InterpDetailsMovingLeastSquaresCoefficients.hpp
```diff
@@ -199,19 +199,34 @@ class MovingLeastSquaresCoefficientsKernel
...
+// to run with a workgroup size of 64.
+#ifdef KOKKOS_ENABLE_HIP
+    if (perTargetMem() <
+        Kokkos::TeamPolicy<ExecutionSpace>::scratch_size_max(1))
+      Kokkos::abort("Can't allocate enough scratch space!");
+    _scratch_level = 1;
+#else
+    if (perTargetMem() <
+        Kokkos::TeamPolicy<ExecutionSpace>::scratch_size_max(0))
+      _scratch_level = 0;
```
aprokop: Would this check be a problem for HIP? Do we not ever want to run scratch 0 for HIP?

src/interpolation/detail/ArborX_InterpDetailsMovingLeastSquaresCoefficients.hpp
```diff
@@ -199,19 +199,34 @@ class MovingLeastSquaresCoefficientsKernel
...
-    Kokkos::TeamPolicy dummy_policy(space, 1, Kokkos::AUTO);
-    dummy_policy.set_scratch_size(0, Kokkos::PerThread(perTargetMem()));
+// FIXME_HIP The HIP backend is limited by the small level 0 scratch space since
+// Kokkos through at least version 5.2 requires at least enough scratch memory
+// to run with a workgroup size of 64.
+#ifdef KOKKOS_ENABLE_HIP
+    if (perTargetMem() <
+        Kokkos::TeamPolicy<ExecutionSpace>::scratch_size_max(1))
+      Kokkos::abort("Can't allocate enough scratch space!");
+    _scratch_level = 1;
```
aprokop: Why not move the determination of the `_scratch_level` into the constructor, and make it non-mutable?

src/interpolation/detail/ArborX_InterpDetailsMovingLeastSquaresCoefficients.hpp
masterleinad: No, the check isn't the problem. The issue is that calling `team_size_recommended` might fail as we have seen in the unit tests. This might be a problem with other backends as well but since we haven't observed that, I just treat HIP special here (also because Vertex-CFD is struggling with this). Ideally we would just move forward with https://github.com/kokkos/kokkos/pull/9339.

src/interpolation/detail/ArborX_InterpDetailsMovingLeastSquaresCoefficients.hpp
masterleinad: We could move all of this into the constructor.

src/interpolation/detail/ArborX_InterpDetailsMovingLeastSquaresCoefficients.hpp
```diff
@@ -199,19 +199,34 @@ class MovingLeastSquaresCoefficientsKernel
   }
 
   auto makePolicy(ExecutionSpace const &space) const
+
```
masterleinad:
```suggestion
```

src/interpolation/detail/ArborX_InterpDetailsMovingLeastSquaresCoefficients.hpp
aprokop, line 91: How about
```C++
    if (perTargetMem() >
        static_cast<std::size_t>(
            Kokkos::TeamPolicy<ExecutionSpace>::scratch_size_max(1)))
      Kokkos::abort("Can't allocate enough scratch space!");

#ifdef KOKKOS_ENABLE_HIP
    // FIXME_HIP The HIP backend is limited by the small level 0 scratch space since
    // Kokkos through at least version 5.2 requires at least enough scratch memory
    // to run with a workgroup size of 64.
    _scratch_level = 1;
#else
    std::size_t scratch_0_max_size = Kokkos::TeamPolicy<ExecutionSpace>::scratch_size_max(0);
    _scratch_level = (perTargetMem() > scratch_0_max_size);
#endif
```

masterleinad: Sure, that's fine with me.

aprokop (approved):
LGTM, but need to wait for CI to pass.

masterleinad:
Retest this please

masterleinad:
The HIP build went through fine.

PR #1388, opened 2026-08-27
Move standard geometric functions out of Ray

aprokop:


dalg24 (commented):
I think the original intent was that the user "opt in" and include the Ray header if he wants to use it, but that it is not considered a basic core geometry.
I don't necessarily have a problem with these changes but I would like you to discuss the rationale for that change.

